PR #17, opened 2019-05-30
add location_code and ServiceLocations

afomi:


afomi:
not proceeding with this branch at this time

PR #95, opened 2019-11-05
Github-auth

aktfrikshun:
Added github omniauth authorization in place of login.gov when git client id is set.

Dockerized local dev so there are no requirements on host os to build developer env other than docker and git.

Steps:
1> clone touchpoints
2> cd touchpoints
3> checkout github-auth branch
4> build docker images: "docker-compose up --build"
5> run db:create: "docker-compose exec website rails db:create"
6> run db:setup: "docker-compose exec website rails db:setup"
7> launch app at http://localhost:3000 and attempt to signin with github
8> You will get an org not found error. You will need to login to the rails console and find your github user and set his organization_id and admin flag. Then all will be good to run locally.


app/controllers/site_controller.rb
```diff
@@ -1,5 +1,6 @@
 class SiteController < ApplicationController
   def index
+    Rails.logger.debug("** AKT ** index")
```
afomi: rm

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -7,6 +7,14 @@ def login_dot_gov
     login
   end
 
+  def github
+    Rails.logger.debug("** AKT ** login_github")
```
afomi: cleanup

app/models/user.rb
```diff
@@ -39,6 +39,8 @@ def self.from_omniauth(auth)
   end
 
   def tld_check
+    return true if ENV['GITHUB_CLIENT_ID'].present?
```
afomi: let's revisit, as this doesn't allow `ensure_organization` to run, later in this method

config/database.yml
```diff
@@ -23,6 +23,8 @@ default: &default
 
 development:
   <<: *default
+  host: postgres
```
afomi: postgresql

afomi:
merging, although the docker stuff needs a bit more work to run fluently. i've noted a card for that.

PR #102, opened 2019-11-09
Touchpoints-719 set provider kind rather than hardcoding to Login.gov…

aktfrikshun:
Touchpoints-719 set provider kind rather than hardcoding to Login.gov for authentication success and failure messages

If route is login_github set kind to "GitHub"

if route is login_dot_gov set kind to "Login.gov"

Use kind in failure and success messages.

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -1,5 +1,6 @@
 class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def login_dot_gov
+    @kind = "Login.gov"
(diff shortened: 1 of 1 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -1,5 +1,6 @@
 class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def login_dot_gov
+    @kind = "Login.gov"
```
afomi: I looked at `auth_hash["provider"]` for this info, but it returns `github` and `login_dot_gov`. I think `Login.gov` and `GitHub` is better, per what you have. 👍

Issue #52, opened 2019-08-01
check Cloud Foundry manifests into version control


afeld:
Rather than only including [a sample](https://github.com/GSA/touchpoints/blob/master/manifest.sample.yml), can we commit the real files, taking the environment variables out? That will also mean that you aren't keeping those credentials locally in a file, which would be a good change.

afomi:
https://github.com/GSA/touchpoints/compare/develop...manifest?expand=1

afeld:
I suggest putting as many of the non-default [attributes](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html) in there as you can, such as the [routes](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html#routes).

afomi:
@afeld - I've added the routes to the cloud.gov instances.

Would you like the cloud.gov services referenced as well?

afeld:
The more the better!

afomi:
@afeld - how's https://github.com/GSA/touchpoints/pull/110?

afomi:
Manifests for Staging, Demo, and Prod have now been included.

https://github.com/GSA/touchpoints/tree/develop

PR #110, opened 2019-11-15
check in manifests for each env

afomi:
https://github.com/GSA/touchpoints/issues/52

### DIFF
--- a/touchpoints-demo.yml
+++ b/touchpoints-demo.yml
@@ -1,0 +1,31 @@
+applications:
+  - name: touchpoints-demo
+    command: bundle exec rake cf:on_first_instance db:migrate && bundle exec rails s -b 0.0.0.0 -p $PORT -e $RAILS_ENV
+    env:
+      AWS_ACCESS_KEY_ID:
+      AWS_REGION:
+      AWS_SECRET_ACCESS_KEY:
+      LOGIN_GOV_CLIENT_ID: urn:gov:gsa:openidconnect.profiles:sp:sso:gsa-tts-opp:touchpoints-demo
+      LOGIN_GOV_IDP_BASE_URL: https://idp.int.identitysandbox.gov/
+      LOGIN_GOV_PRIVATE_KEY:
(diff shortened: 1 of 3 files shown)


touchpoints-demo.yml
```diff
@@ -0,0 +1,31 @@
...
+      S3_AWS_BUCKET_NAME:
+      S3_AWS_HOST:
+      S3_AWS_REGION:
+      S3_AWS_SECRET_ACCESS_KEY:
+      TOUCHPOINTS_EMAIL_SENDER:
+      TOUCHPOINTS_GTM_CONTAINER_ID:
+      TOUCHPOINTS_WEB_DOMAIN: touchpoints-demo.app.cloud.gov
+    buildpacks:
+      - https://github.com/cloudfoundry/nodejs-buildpack.git
+      - https://github.com/cloudfoundry/ruby-buildpack.git
```
afeld: I recommend using the cloud.gov-provided ones if you can: `nodejs_buildpack` and `ruby_buildpack`. They have more compliance baked in than the public ones.

touchpoints-demo.yml
```diff
@@ -0,0 +1,31 @@
...
+      LOGIN_GOV_REDIRECT_URI: https://touchpoints-demo.app.cloud.gov/users/auth/login_dot_gov/callback
+      NEW_RELIC_KEY:
+      RAILS_ENV: production
+      S3_AWS_ACCESS_KEY_ID:
+      S3_AWS_BUCKET_NAME:
+      S3_AWS_HOST:
+      S3_AWS_REGION:
+      S3_AWS_SECRET_ACCESS_KEY:
+      TOUCHPOINTS_EMAIL_SENDER:
+      TOUCHPOINTS_GTM_CONTAINER_ID:
```
afeld: You could probably leave out the environment variables whose values aren't in this file, if you want. Fine either way.

afomi: will leave them in for now. i might reference the manifest.sample later with documentation to ensure a new dev picks these up.   leaving them here sorta prompts the config.

thanks!

touchpoints-demo.yml
afomi: nice.  will do. thanks!

PR #121, opened 2019-11-22
TP-766 First pass at custom 404 and 500 pages

aktfrikshun:
TP-766 ( TP-756 ) First pass at custom 404 and 500 pages.


app/views/errors/internal_server_error.html.erb
```diff
@@ -0,0 +1,61 @@
+<!DOCTYPE html>
```
afomi: could re-use 'public' layout

config/routes.rb
```diff
@@ -1,6 +1,10 @@
 Rails.application.routes.draw do
   devise_for :users, controllers: { omniauth_callbacks: 'users/omniauth_callbacks' }
 
+
+  match "/404", :to => "errors#not_found", :via => :all
```
afomi: 👍

PR #131, opened 2019-12-04
TP-788 Hide character counters if no character limit has been set by …

aktfrikshun:
TP-788 Hide character counters if no character limit has been set by admin.

This simply hides the counter fields if the admin as not set a limit.  The client side javascript to enforce limits is still in place as is the max limits set in HTML on the text fields.


afomi:
LGTM!

PR #136, opened 2019-12-10
TP-544 Deactivate endpoint, notification and event logging

aktfrikshun:
TP-544 Deactivate endpoint, notification and event logging


app/models/event.rb
```diff
@@ -0,0 +1,25 @@
+class Event < ApplicationRecord
+
+	# Extend this list with all First Class event types to be logged TP-
+	@@names = {
```
afomi: whitespace

aktfrikshun: I added a rubicon format on save to my IDE.   Seems like it won't indent the methods following a "private".

I added a check for request_source_authorized? which currently simply checks for the LOGIN_GOV_PRIVATE_KEY value in the header matching the ENV variable.  We can substitute the real authorization implementation when it becomes known.

Also added a couple tests to confirm the appropriate response is sent on deactivate in the authorized and unauthorized cases.

PR #139, opened 2019-12-11
remove Services

afomi:
* still need to remove Service fields from the data model in a future pass


app/controllers/admin/forms_controller.rb
```diff
@@ -1,7 +1,6 @@
 class Admin::FormsController < AdminController
   before_action :set_form, only: [:show, :edit, :update, :copy, :destroy]
   before_action :set_touchpoint, only: [:show, :edit]
-  before_action :ensure_organization_manager
 
   def index
     @forms = Form.all.order("name ASC")
```
afomi: more work needs to be done to scope Forms to users who can see them. 

either, 1) Users are associated to a Form when creating a Form.  2) Form is associated with a Touchpoint.

afomi: this is also because the `ensure_organization_manager` is removed.

More users should be able to access Forms.

PR #146, opened 2019-12-16
TP-701 - Add state machine for states of a touchpoint

aktfrikshun:
Levarage aasm gem

### DIFF
--- a/app/models/touchpoint.rb
+++ b/app/models/touchpoint.rb
@@ -1,4 +1,6 @@
 class Touchpoint < ApplicationRecord
+  include AASM
+
   belongs_to :form, optional: true
   belongs_to :organization
   has_many :submissions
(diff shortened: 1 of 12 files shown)


aktfrikshun:
Currently allowing for all states to be transitioned to from any state.  Simply need to change the event state definitions to restrict transitionable states.

Options for select on Touchpoint edit view are restricted to states transitionable to from the current state.

app/models/touchpoint.rb
```diff
@@ -31,14 +35,48 @@ def omb_number_with_expiration_date
...
+
   scope :active, -> { where("id > 0") } # TODO: make this sample scope more intelligent/meaningful
 
   def send_notifications?
     self.notification_emails.present?
   end
 
   def deployable_touchpoint?
-    self.form ? true : false
+    self.live? ? true : false
```
afomi: `self.live?` is equivalent

app/models/touchpoint.rb
```diff
@@ -31,14 +35,48 @@ def omb_number_with_expiration_date
     "inline"
   ]
 
+
+  aasm do
```
afomi: 👍

PR #149, opened 2019-12-23
TP-538 Add sidekiq and send email async

aktfrikshun:
https://github.com/GSA/touchpoints/pull/new/TP-538


app/assets/javascripts/file-saver/CHANGELOG.md
```diff
@@ -0,0 +1,60 @@
+# Changelog
```
afomi: nm

aktfrikshun: Do you have redis/sidekiq running locally?

PR #167, opened 2020-01-28
WIP: TP-820 Inline insertion of form components while form building

afomi:


afomi:
@aktfrikshun - the single-page form builder improvements provide stand alone value. Consider getting this branch in on the way to the bigger Edit page redesign.

aktfrikshun:
@afomi Will do :-)

PR #176, opened 2020-02-06
Touchpoint to form

afomi:
Sharing WIP on branch that merges Touchpoints with Forms.

Combining the Touchpoint and Form object is a major change that affects most parts of the app.
The result should be a simplified user experience, where a user just manages a Form and its workflow, rather than the additional Touchpoint object that is used to hold most of the PRA-related workflow information.


afomi:
@aktfrikshun - this work entails removing the `Touchpoint` model altogether, and migrating its attributes to its `form`, which is has a 1:1 relationship with.

I can see future work refactoring out PRA and non form-specific attributes.

PR #218, opened 2020-04-13
Updated all actions of question to stay on edit form

kprabhav:
Current action on adding/editing/adding question takes to main form instead of staying on edit form. This is updated to stay on the edit form upon successful completion of the actions.

TODO: tests


afomi:
going to use https://github.com/GSA/touchpoints/pull/227 instead

PR #235, opened 2020-05-27
WIP: adding nonces to inline javascripts

ryanwoldatwork:
* based on https://github.com/18f/identity-idp's implementation of secure_headers


ryanwoldatwork:
Touchpoints supports a sha-256 integrity hash for each form's payload.

However, it is a bit brittle in the sense that if any element on the form changes, the integrity hash is invalidated.
This makes it brittle for adjustments, and subject to breakage if there are underlying changes to the form's markup, or USWDS updates, for example.

PR #243, opened 2020-06-11
Release 0.41.0

ryanwoldatwork:
* Form supports logo upload (in lieu of the logo on Organization)
* Distinguish between Forms and FormTemplates in queries
* Set UUID for each Submission response
* Set REDIS and S3 from VCAP_SERVICES
* Default Form.load_css to true (users can toggle it off after creation)
* Publish test coverage in `/coverage` using SimpleCov


ryanwoldatwork:
Will need to remove the REDIS work before merge, as the update REDIS version is still in Cloud.gov Beta.

PR #255, opened 2020-07-29
improve accessibility of modal tab

ryanwoldatwork:
* use the .usa-button style
* adjust the fixed, bottom-right positioning


ryanwoldatwork:
the existing tab button:

existing-tab-button

the proposed tab button:

proposed-tab-button

ryanwoldatwork:
a skiplink `<a href>` has been added to the beginning of `<body>`. It has an ID of the floating modal button. It now has an event listener to place `focus()` on the floating modal button when clicked.

This gets a keyboard there in 3 clicks:

1. skip to main (tab)
1. skip to feedback (select)
1. floating feedback button (select) -- to open the modal with the form

PR #264, opened 2020-08-18
TP-1061 Single Page Form Builder - WIP

aktfrikshun:


ryanwoldatwork:
Looking good!

Notes:

To address before merge:

* Section titles don't seem to update in place
* Move questions up/down doesn't seem to work
* The modal treatment with the x is a bit "tabby". I'd prefer to use our standard modal with the "x", like on https://touchpoints.digital.gov's modal.
* Add tests for these

Could follow up on:

* Gray styling is heavy and does not match style wise
* More than 2 Hover states for the 3 dots show when activated. Hovering over a new one should close others.
* Move the affordance for a new question to the Section and move it out of the 3 dots

PR #288, opened 2020-09-09
list inactive users

ryanwoldatwork:
automate the report of users who haven't logged in in 90 days

### DIFF
--- a/spec/features/admin/users_spec.rb
+++ b/spec/features/admin/users_spec.rb
@@ -51,6 +51,18 @@ feature "Managing Users", js: true do
           expect(page).to have_link(form.name)
         end
       end
+
+      describe "/admin/users/inactive" do
+        before do
+          visit inactive_admin_users_path
+        end
+
+        it "displays inactive users" do
+          expect(page).to have_css("table")
+          expect(User.count).to eq(2) # out of 2
+          expect(page).to have_content(organization_manager.email) # only 1 is inactive
+        end
+      end
     end
 
     describe "edit a User" do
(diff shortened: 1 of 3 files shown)


spec/features/admin/users_spec.rb
```diff
@@ -72,22 +84,37 @@
...
+      it "does not display Admin nor Last Sign in at" do
+        expect(page).not_to have_content("Admin")
+        expect(page).not_to have_content("Last signed in")
       end
     end
 
-    it "does not display Admin nor Last Sign in at" do
-      expect(page).not_to have_content("Admin")
-      expect(page).not_to have_content("Last signed in")
+    describe "/admin/users/inactive" do
```
ryanwoldatwork, line 110: this is the new code

spec/features/admin/users_spec.rb
```diff
@@ -72,22 +84,37 @@
...
     end
 
-    it "permitted to see the page" do
-      expect(page).to have_content("Users")
-      within "table" do
-        expect(page).to have_content("Org mgr")
-        expect(page).to have_content("Email")
-        expect(page).to have_content("Organization Name")
-        expect(page).to have_link("Edit")
+    describe "admin" do
```
ryanwoldatwork, line 89: this is just indented into a describe block

aktfrikshun (approved):
I don't see any issues here.



We may want create a scheduled job that marks a user as inactive if they haven't logged in in 90 days ( in addition to other criteria we might add in the future for inactivating a user ).

ryanwoldatwork:
I agree a job could improve this automation. Communication to the client is needed somewhere in that flow. TBD.

PR #291, opened 2020-09-11
TP-1081 Drag questions across sections

aktfrikshun:
TP-1081 Drag questions across sections

Works, but UI might be better if we provided a "Move" icon and used it as a drag handle.  It' might be confusing as to what part of the UI is the section and which is individual questions when dragging.

### DIFF
--- a/app/assets/stylesheets/site.scss
+++ b/app/assets/stylesheets/site.scss
@@ -492,5 +492,5 @@ a.form-section-save {
   width: 400px;
 }
 .questions {
-  min-height: 100px;
+  min-height: 10px;
 }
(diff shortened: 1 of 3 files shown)


ryanwoldatwork:
The questions are dragging between Form Sections 👍 

Can the drop area be placed above the "Add Question" and "Delete Section" buttons?

dragarea

aktfrikshun:
@ryanwoldatwork   Latest commit on PR https://github.com/GSA/touchpoints/pull/291 has the questions drop area  separate from and above the section buttons

ryanwoldatwork (approved):

PR #295, opened 2020-09-15
TP-1079 1st attempt at using the uswds datepicker widget

aktfrikshun:
TP-1079 1st attempt at using the uswds datepicker widget

@ryanwoldatwork It doesn't appear all the uswds image assets have been integrated into the project.  The calendar icon in particular is missing.  I've added it to assets/images but am getting the following error when trying to render:

This page contains the following errors:
error on line 51 at column 89: Specification mandates value for attribute data-pjax-transient
Below is a rendering of the page up to the first error.
uswds/calendar-alt-solid.svg at develop · uswds/uswds

I'm also having trouble getting an existing value to display in the date text field.   I started with the form helper methods and wound up writing the HTML manually to match the uswds component html.   No joy.  Thanks for taking a look.


ryanwoldatwork:
@aktfrikshun - I updated the USWDS library to 2.8.1 and updated the svgs for the calendar, left and right arrows

ryanwoldatwork (approved):
this should be ready to go shortly

PR #296, opened 2020-09-16
TP-1073 integrate uswds character counter widget in survey textarea …

aktfrikshun:
TP-1073 integrate uswds character counter widget in survey textarea and textfield question types

### DIFF
--- a/app/views/components/forms/question_types/_text_field.html.erb
+++ b/app/views/components/forms/question_types/_text_field.html.erb
@@ -1,9 +1,15 @@
 <div class="radios" role="group" aria-labelledby="<%= question.answer_field %>">
   <%= render 'components/question_title', question: question, question_number: question_number %>
   <% if question.character_limit && question.character_limit > 0 %>
-    <%= text_field_tag question.answer_field.to_sym, nil, class: "usa-input", required: question.is_required, maxlength: question.max_length, onkeyup: "touchpointForm#{form.short_uuid}.textCounter(this,#{question.max_length});" %>
+  	<div class="usa-character-count">
+	    <div class="usa-form-group">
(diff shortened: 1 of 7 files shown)


app/views/components/forms/question_types/_text_field.html.erb
```diff
@@ -1,9 +1,15 @@
 <div class="radios" role="group" aria-labelledby="<%= question.answer_field %>">
   <%= render 'components/question_title', question: question, question_number: question_number %>
   <% if question.character_limit && question.character_limit > 0 %>
-    <%= text_field_tag question.answer_field.to_sym, nil, class: "usa-input", required: question.is_required, maxlength: question.max_length, onkeyup: "touchpointForm#{form.short_uuid}.textCounter(this,#{question.max_length});" %>
+  	<div class="usa-character-count">
+	    <div class="usa-form-group">
```
ryanwoldatwork, line 5: tab / spacing?

ryanwoldatwork:
The `/example` path doesn't seem to update.

Screen Shot 2020-09-17 at 3 43 08 PM

aktfrikshun:
> The `/example` path doesn't seem to update.
> 
> Screen Shot 2020-09-17 at 3 43 08 PM

@ryanwoldatwork It does not look to me like uswds.js is being loaded.   In the hosted form it is loaded as part of application.js.  I have tried to include it in the demo page as it uses it's own layout, but am not having much luck.

ryanwoldatwork:
hi @aktfrikshun - if only the CSS formatting, markup, and text are updated, that'd be good.

The prior .js worked and you're right, we can't rely on the uswds.js to be loaded, as some clients may use it.

PR #297, opened 2020-09-17
TP-717 TP-717 Persist checkbox values to db as comma separated string

aktfrikshun:
TP-717 Persist checkbox values to db as comma separated string

@ryanwoldatwork this ones ready to go with specs


ryanwoldatwork (commented):
Check the `question_type_javascript_params` method in `application_helper.rb`, which selects the checkbox inputs.

PR #304, opened 2020-09-23
TP-1085 support inline edit of question option text, add inline delete

aktfrikshun:
TP-1085 support inline edit of question option text, add inline delete

Hey Ryan, pay close attention to the UI flow here.  I'd like your recommendation.   

This is the first UI experience that has inline edit with no edit or save button.  You just click on the text and hit enter to save.

I left the Edit link ( but changed it to a pencil ) so we can update the option value and position.   That feels a little wonky to me as there is a lot of edit options going on here.

Inline delete works as expected, we immediately remove the node from the dom and submit an ajax request to delete the option.

Just play around with the flow and let me know of any recommended changes.


ryanwoldatwork (approved):
@aktfrikshun - i think this is ready to go

PR #323, opened 2020-10-14
TP-1022 Enhanced user time_zone handling

aktfrikshun:
TP-1022 Enhanced user time_zone handling

User sets time_zone preference in user profile and in survey will be two follow on tickets.


app/views/admin/submissions/index.html.erb
```diff
@@ -44,7 +44,7 @@
             <%= submission.referer %>
           </td>
           <% end %>
-          <td><%= submission.created_at %></td>
+          <td><%= format_time(submission.created_at, submission.form.time_zone.present? ? submission.form.time_zone : current_user.time_zone) %></td>
```
ryanwoldatwork, line 47: viewing submissions by MY timezone vs the Form's timezone is something to consider, UX-wise

ryanwoldatwork (approved):

PR #330, opened 2020-10-26
TP-981 default route is now site#index which redirects based on ENV["INDEX_URL"]

aktfrikshun:
TP-981 default route is now site#index which redirects based on ENV["INDEX_URL"]


ryanwoldatwork:
envs have been updated with `INDEX_URL`

PR #338, opened 2020-11-03
TP-1122 Add API-KEY to user profile with ability to generate/remove it

aktfrikshun:
TP-1122 Add API-KEY to user profile with ability to generate/remove it


lib/api_key.rb
```diff
@@ -0,0 +1,7 @@
+# lib/api_key.rb
+
+module ApiKey
+  def self.generator
+    SecureRandom.base64.tr('+/=', 'Qrt')
```
ryanwoldatwork, line 5: good to start. 
this will be replaced in a future iteration.

PR #339, opened 2020-11-03
TP-1134 Add OTHER logic to radio buttons

aktfrikshun:
TP-1134 Add OTHER logic to radio buttons

### DIFF
--- a/app/views/components/forms/question_types/_checkbox.html.erb
+++ b/app/views/components/forms/question_types/_checkbox.html.erb
@@ -11,7 +11,7 @@
         <%= option.text %>
       <% end %>
       <% if ["OTHER", "OTRO"].include?(option.text.upcase) %>
-        <input type="text" name="<%= question.answer_field %>_other" id="<%= question.answer_field %>_other" data-checkbox-id="<%= @option_id %>" placeholder="<%= t 'form.enter_other_text' %>" class="usa-input other-option" />
+        <input type="text" name="<%= question.answer_field %>_other" id="<%= question.answer_field %>_other" data-option-id="<%= @option_id %>" placeholder="<%= t 'form.enter_other_text' %>" class="usa-input other-option" />
         <br/>
       <% end %>
     </div>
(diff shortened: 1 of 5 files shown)


app/views/components/forms/question_types/_radio_buttons.html.erb
```diff
@@ -9,6 +9,10 @@
       <%= label_tag(nil, for: "#{question.answer_field}_#{(index + 1).to_s}", class: "usa-radio__label") do %>
         <%= option.text %>
       <% end %>
+      <% if ["OTHER", "OTRO"].include?(option.text.upcase) %>
```
ryanwoldatwork, line 12: this works if the 'other' radio is selected, then the value will be 'other', even if there is text in the text field.

the value of the response should be the typed in value, or if blank, then the value of the `other` text

ryanwoldatwork: and, got specs? 

🔴  🟢

aktfrikshun: Confirmed radio button works with specs

app/views/components/forms/question_types/_checkbox.html.erb
```diff
@@ -11,7 +11,7 @@
         <%= option.text %>
       <% end %>
       <% if ["OTHER", "OTRO"].include?(option.text.upcase) %>
-        <input type="text" name="<%= question.answer_field %>_other" id="<%= question.answer_field %>_other" data-checkbox-id="<%= @option_id %>" placeholder="<%= t 'form.enter_other_text' %>" class="usa-input other-option" />
```
ryanwoldatwork, line 14: 👍

spec/factories/form.rb
```diff
@@ -180,5 +180,19 @@
...
     end
 
+    trait :radio_button_form do
+      name { "Radio button form" }
+      kind { "custom" }
+      after(:create) do |f, evaluator|
+        FactoryBot.create(:question,
+          :with_radio_buttons,
+          form: f,
+          answer_field: :answer_03,
```
ryanwoldatwork, line 190: Why `_03` instead of `_01`?

PR #342, opened 2020-11-05
TP-1123 Response pagination 

aktfrikshun:
This is first draft, no ajax
AJAX coming soon


app/assets/stylesheets/site.scss
```diff
@@ -489,3 +489,9 @@ a.form-section-save {
   border: none;
   border-bottom: 1px solid black;
 }
+nav {
```
ryanwoldatwork, line 492: Is this style intended to be scoped for the pagination?
I'm assuming `nav` is used elsewhere.

app/models/submission.rb
```diff
@@ -9,6 +9,9 @@ class Submission < ApplicationRecord
 
   scope :non_flagged, -> { where(flagged: false) }
 
+  paginates_per 10
```
ryanwoldatwork, line 12: Sync with Natalie on this, but 100 at a time feels more reasonable than 10.

I'm thinking about if sorting submissions by `created_at DESC` would make more sense here than `ASC`. 
My thought is about having thousands of responses, and how to do I see latest easily? 
* look at first page of `DESC` sorted results
* click to last page and view `ASC` sorted results

ryanwoldatwork (changes requested):

aktfrikshun:
Makes sense on all points.  Ill have the ajax flavor soon.

ryanwoldatwork (approved):

PR #343, opened 2020-11-05
remove organization manager

ryanwoldatwork:
* form manager permissions are sufficient
* organization-wide permissions are too broad and not useful


aktfrikshun (approved):
Very thorough job stripping out Org Manager.  Nice work!

PR #347, opened 2020-11-11
TP-1137 Add api_key_udpate_date and warn if user api_key has not been…

aktfrikshun:
TP-1137 Add api_key_udpate_date and warn if user api_key has not been updated in the past 6 months


app/models/user.rb
```diff
@@ -109,11 +109,11 @@ def self.to_csv
   end
 
   def set_api_key
-    update(api_key: ApiKey.generator)
+    update(api_key: ApiKey.generator, api_key_update_date: Time.now)
```
ryanwoldatwork, line 112: `_updated_at` is a bit more conventional, per Rails

spec/models/user_spec.rb
```diff
@@ -90,6 +90,31 @@
         expect(@user.errors.messages[:organization].first).to include("'toomany.associates.subdomain.example.gov' has not yet been configured for Touchpoints")
       end
     end
+
+    context "api key" do
```
ryanwoldatwork, line 94: 👍

PR #355, opened 2020-11-19
V0 api

ryanwoldatwork:


ryanwoldatwork:
@aktfrikshun - this branch uses /v0 instead of /v1 as a prefix for the API, because the API is in `beta` and has not completed API review by GSA IT.

This branch also changes the key generation from a random local generator, to allowing a user to copy/paste their key in for now.  The api.data.gov gateway will validate the key, and then the key must still match an existing user.

aktfrikshun (approved):

PR #360, opened 2020-11-25
TP-1144 Implicit save of survey title, instructions & disclaimer

aktfrikshun:
TP-1144 Implicit save of survey title, instructions & disclaimer

@ryanwoldatwork I've changed the inputs to spans/divs with contenteditable and tabindex.

You should now be able to tab into each element and update it's value.  I set the tab order to start with the title, then instructions, then all sections then the disclaimer text.

Tab around and let me know what you think of this approach.


ryanwoldatwork:
Hi @aktfrikshun - let's pause on this for a bit and sync with Design.

Goals here include:
* accessible
* pleasant
* reduce the distance between Preview and Edit states

I think bullets 2 and 3 are improved, but we need to ensure an accessible approach. Hopefully there are pleasant designs that don't require basic input text boxes.  If not, then, the text boxes here will have to do for now.

aktfrikshun:
> Hi @aktfrikshun - let's pause on this for a bit and sync with Design.
> 
> Goals here include:
> 
> * accessible
> * pleasant
> * reduce the distance between Preview and Edit states
> 
> I think bullets 2 and 3 are improved, but we need to ensure an accessible approach. Hopefully there are pleasant designs that don't require basic input text boxes. If not, then, the text boxes here will have to do for now.

I removed the input and text boxes in favor of spans and divs with contenteditable.  I think it is visually appealing and accessible .  Let's discuss on standup

ryanwoldatwork (commented):
This version looks much better.

* [ ] I got into a situation where Form Section was blank (set to "") and could not edit it.
* [ ] Is it possible to match the display styles more closely? So the user sees as clearly as possible how the form will be displayed?

All in all, feels much cleaner.

PR #362, opened 2020-12-01
TP-1142 add response_count, last_response_created_at to form, update …

aktfrikshun:
P-1142 add response_count, last_response_created_at to form, update on submission create

These fields will be set appropriately whenever a submission is created, however we could create a rake task to update all existing surveys that may no longer be active

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -46,8 +46,9 @@ class SubmissionsController < ApplicationController
     @submission.referer = submission_params[:referer]
     @submission.page = submission_params[:page]
     @submission.ip_address = request.remote_ip
-
-    create_in_local_database(@submission)
+    @result = create_in_local_database(@submission)
+    @form.update(response_count: @form.submissions.size, last_response_created_at: @submission.created_at)
+    @result
   end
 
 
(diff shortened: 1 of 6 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -46,8 +46,9 @@ def create
     @submission.referer = submission_params[:referer]
     @submission.page = submission_params[:page]
     @submission.ip_address = request.remote_ip
-
-    create_in_local_database(@submission)
+    @result = create_in_local_database(@submission)
+    @form.update(response_count: @form.submissions.size, last_response_created_at: @submission.created_at)
```
ryanwoldatwork, line 50: how about doing this in a callback on `Form.rb` - `before_update`

app/views/admin/forms/show.html.erb
```diff
@@ -240,4 +240,12 @@
       </small>
     </p>
   </div>
+  <div>
+    <p>
+      <strong>Number of responses:</strong> <%= @form.response_count %>
+      <% if @form.last_response_created_at.present? %>
+        ( <small>last response created on <%= format_time(@form.last_response_created_at, current_user.time_zone) %></small> )
```
ryanwoldatwork, line 247: 👍

app/controllers/submissions_controller.rb
ryanwoldatwork: Check out `counter_cache` https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to

PR #363, opened 2020-12-02
TP-1147 Add remote partials and js for submission flag/unflag

aktfrikshun:
TP-1147 Add remote partials and js for submission flag/unflag

### DIFF
--- a/app/views/admin/submissions/flag.js.erb
+++ b/app/views/admin/submissions/flag.js.erb
@@ -1,1 +1,1 @@
-$("#flag-submission-<%= @submission.id %>").parent().html("<%= j (render partial: 'flag', :locals => {submission: @submission}) %>")
+$('#td-flag-<%= @submission.id %>').html("<%= j (render partial: 'flag', :locals => {submission: @submission}) %>")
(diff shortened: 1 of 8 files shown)


spec/features/admin/submissions_spec.rb
```diff
@@ -47,7 +47,6 @@
             end
 
             it "successfully flags Submission" do
-              expect(page).to have_content("Response #{submission.id} was successfully flagged.")
```
ryanwoldatwork, line 50: is there a way to assert the flag happened? maybe by flag color/status?

app/views/admin/submissions/flag.js.erb
```diff
@@ -0,0 +1 @@
+$("#flag-submission-<%= @submission.id %>").parent().html("<%= j (render partial: 'flag', :locals => {submission: @submission}) %>")
```
ryanwoldatwork, line 1: can the element be targeted by submission id, then within the submission, find the flag td/element?

spec/features/admin/submissions_spec.rb
aktfrikshun: Page having content Flagged asserts the flag happened

app/views/admin/submissions/flag.js.erb
aktfrikshun: Sure, I can do that

aktfrikshun: @ryanwoldatwork  Done

PR #369, opened 2020-12-07
ensure response_count is used

ryanwoldatwork:
* instead of submissions.size, which queries the DB

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -157,6 +157,7 @@ class Form < ApplicationRecord
     new_form.name = "Copy of #{self.name}"
     new_form.title = new_form.name
     new_form.survey_form_activations = 0
+    new_form.response_count = 0
     new_form.aasm_state = :in_development
     new_form.uuid = nil
     new_form.legacy_touchpoint_id = nil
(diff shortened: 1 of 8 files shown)


aktfrikshun (commented):
@ryanwoldatwork According to the documentation for counter_cache option on ActiveRecord, you do not have to reference response count directly.   

Simply using the access form.submissions.size will use the cached value.

In fact, you can call response_count anything and you won't have to update the entire code base wherever form.submissions.size is used.

ryanwoldatwork:
thanks @aktfrikshun

I saw something like that as well, but the logs show a different story, so I went forth with these changes.

submission-queries

I think the specific issue is that the field needs to be named `submissions_count` in order for this to work.

aktfrikshun:
I’d be curious to see if the query was actually hitting the db or if the SQL is being intercepted by the helper. 

The docs specifically say you can call the counter field anything you want so long as it is named in the belongs_to association.

But it probably makes no difference as we won’t be changing the name of the counter field.   It would just be nice if we could rely on the cache to work as expected if future developers reference the collection size method unaware.

Allen

ryanwoldatwork:
I tested this locally using `submissions_count` and it seems to work, meaning the SQL query isn't called.

In the future, `response_count` could be renamed to `submissions_count`

PR #372, opened 2020-12-12
add location_code to .csv export

ryanwoldatwork:


aktfrikshun (commented):
I see this adds the location_code to the export hash.   Do we need to prove that if a location code exists it is returned in export?  I only see where we are verifying the header key exists.  I'm not too familiar with how this piece works.

ryanwoldatwork:
This change adds the location_code to the .csv export. Not conditionally, but always.

PR #382, opened 2021-01-28
Converting api json responses to json-api format per GSA guidelines

ryanwoldatwork:


aktfrikshun (approved):
Looks good.   I've used the Serializer pattern before.  It's nice to be able to explicitly specify data attributes included.

PR #384, opened 2021-02-04
TP-1170 cancelling out of create question option form hides the form

aktfrikshun:
TP-1170 Cancelling out of create question option form simply hides the form rather than refreshing the entire page


ryanwoldatwork (approved):
👍

PR #418, opened 2021-04-01
Service stages

ryanwoldatwork:
As a Form Manager,
I want to calculate the burden of a Service,
So that I can support more accurate PRA burden calculations and organizational cost.

As a Form Manager,
I want to be able to report on and share Service Data
So that I can understand and communicate about Services in a higher resolution way.


ryanwoldatwork:
The below are ideas in the prototype phase, shared for feedback.

Here is a list of Services.

Each Service belongs to an Organization in Touchpoints.
A Service has many Service Stages (eg: before, during, after)

service-list

ryanwoldatwork:
Here are basic, default Stages for a Service.

Each Service has many Service Stages.
These Stages represent work and or phases in a defined organizational process related to Service delivery.

service-stages

ryanwoldatwork:
Here is an example BPMN diagram; the most basic version.

This Process has a Start Event, an Intermediate Event, and an End Event. 
Real diagrams would get much more complex than this.

bpmn


#### References

https://bpmn.io/toolkit/bpmn-js/

aktfrikshun (commented):
Question:  Is a barrier within a stage intended to be a workflow checkpoint which must be passed before we can move to the next stage?  If so, I'm assuming will be adding conditional workflow logic for allowing flow between stages along with equity computations triggered on stage state change?

General nav notes:  
Back to surveys doesn't make sense in the whole Services -> Stage -> Barrier flow
When creating a new Barrier on the Service Stage page there is no way to get back to the service stage without clicking through the whole flow again.

PR #425, opened 2021-04-07
TP-1198 User can cancel out of editing form question

aktfrikshun:
Linked ticket: https://trello.com/c/3ZvzJXoM/1198-user-can-cancel-out-of-editing-form-questions

Add cancel on form question.   On cancel it does an AJAX call to show/get rather than update to restore the question view.

Refactored the question view builder to be an independent partial so it can be referenced by show, update and the main form builder rather than duplicating code.

Added specs to test cancel button is present on edit as well as any changes made during the edit view are not persisted.


app/views/admin/questions/_form.html.erb
```diff
@@ -105,6 +106,20 @@
       }
     });
 
+    $('.cancel-question').on("click", function(event) {
```
ryanwoldatwork, line 109: How about something like 

$('.question').on("click", ".cancel-question", function(event) { ... }

This will then allow for getting the dom_id from the `.question` object.

app/views/admin/questions/show.html.erb
```diff
@@ -0,0 +1 @@
+<%= render 'admin/questions/question', { form: @form, question: @question } %>
```
ryanwoldatwork, line 1: good refactor between this, question, and update

ryanwoldatwork (approved):

PR #431, opened 2021-04-15
TP-851 Add ability to invite a user to Touchpoints

aktfrikshun:
TP-851 Add ability to invite a user to Touchpoints

Linked Trello ticket: https://trello.com/c/4K1AeCRJ/851-invite-a-colleague-to-touchpoints

Note:   I did not validate nor provide a spec for .gov / .mil email domains.   There are other domains that may be valid with Github authentication and there may be a use case where the logged in user does not know their colleagues .gov/.mil email address but has another address for them in their contacts.  

When the referred user access the Touchpoints link they will need to register with the appropriate .gov, .mil or Github account  address.

### DIFF
--- a/app/controllers/admin/forms_controller.rb
+++ b/app/controllers/admin/forms_controller.rb
@@ -10,6 +10,7 @@ class Admin::FormsController < AdminController
     :compliance,
     :permissions, :questions, :responses,
     :copy, :copy_by_id,
+    :invite,
     :notifications,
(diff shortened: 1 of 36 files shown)


ryanwoldatwork (approved):
looks good. 

I moved the form from the profile page to the /forms/:id/permissions page. There, the user is prompted to add a user and can add a user who does not exist.

PR #445, opened 2021-04-26
TP-1205 Add email front end validation as well as tests

aktfrikshun:
TP-1205 Add email front end validation as well as tests

Linked Trello ticket: https://trello.com/c/z8yd0gxl/1205-ensure-a-email-text-field-validates-on-the-frontend

### DIFF
--- a/app/views/components/forms/question_types/_text_email_field.html.erb
+++ b/app/views/components/forms/question_types/_text_email_field.html.erb
@@ -9,3 +9,17 @@
     <%= text_field_tag question.answer_field.to_sym, nil, type: :email, class: "usa-input", required: question.is_required, maxlength: question.max_length %>
   <% end %>
 </div>
+<script>
+	$('#<%= question.answer_field.to_sym %>').blur(function() {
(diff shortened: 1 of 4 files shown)


app/views/components/forms/question_types/_text_email_field.html.erb
```diff
@@ -9,3 +9,17 @@
...
   <% end %>
 </div>
+<script>
+	$('#<%= question.answer_field.to_sym %>').blur(function() {
+		var EmailRegex = /^([a-zA-Z0-9_.+-])+\@(([a-zA-Z0-9-])+\.)+([a-zA-Z0-9]{2,4})+$/;
+		var email = $.trim($(this).val());
+		if (email.length == 0) return;
+		result = EmailRegex.test(email);
+		if (!result) {
+			alert('Please enter a valid email address');
```
ryanwoldatwork, line 19: An alert works.

Using the USWDS text styles is where we'll want to go with this soon.
Ensuring errors look okay without USWDS will be a bit tricky.

 https://designsystem.digital.gov/components/text-input/

aktfrikshun: @ryanwoldatwork Good add with the error highlighting.  Between that and the focus I'm sure the user will understand what field needs to be addressed.

PR #452, opened 2021-05-03
TP-1215 Website smart search

aktfrikshun:
TP-1215 Website smart search

Linked Trello ticket:  https://trello.com/c/JAb7IrMp/1215-typeahead-search-for-websites-on-admin-website

@ryanwoldatwork First pass.  I've integrated the new field you added recently.  All displayed data fields will be searched for matching text.   

Not sure I like how the "all/include inactive" works with the search field.

I might suggest simply adding the include inactive as a radio button toggle to further limit search results.  What do you think?

I set a max limit of 50 results with verbiage to help the user to use the search field to narrow results.

I'll add in specs after we agree on behavior -- Thanks!


ryanwoldatwork (approved):
I removed MAX_RESULTS for now.
The search works well and will help sort through 500+ sites.

PR #461, opened 2021-05-12
TP-1220 Capture user url intent when not logged in

aktfrikshun:
TP-1220 Capture user url intent when not logged in

Linked Trello Issue: https://trello.com/c/e4jOOLDH/1220-maintain-url-intent-when-logging-in

Utilizes Devise store user location functionality to save user url intent and then redirect after successful sign in.


ryanwoldatwork (commented):
reviewing

PR #467, opened 2021-05-14
TP-1218 uswds error message and functions to showWarning and showValid

aktfrikshun:
TP-1218 uswds error message and functions to showWarning and showValid


ryanwoldatwork (changes requested):
try entering email `test` and clicking `submit` twice.  The error label is re-flowing the layout in an unintended way.

ryanwoldatwork (approved):

PR #534, opened 2021-07-02
TP-1217 API date filter

aktfrikshun:
TP-1217 API date filter

See note in code code comments:  Defaulting start_date to 1 year ago under the assumption that a user would never need to pull responses over 1 year old via the API.

Linked Trello Ticket: https://trello.com/c/Qb7ABrNt/1217-add-date-filter-to-responses-api

### DIFF
--- a/app/controllers/api/v0/forms_controller.rb
+++ b/app/controllers/api/v0/forms_controller.rb
@@ -8,17 +8,25 @@ class Api::V0::FormsController < ::ApiController
   end
 
   def show
-    @form = current_user.forms.find_by_short_uuid(params[:id])
-    @page_num = (params[:page].present? ? params[:page].to_i : 0)
-    @page_size = (params[:page_size].present? ? params[:page_size].to_i : 500)
-    @page_size = 5000 if @page_size > 5000
+    form = current_user.forms.find_by_short_uuid(params[:id])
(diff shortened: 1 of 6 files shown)


app/controllers/api/v0/forms_controller.rb
```diff
@@ -8,17 +8,25 @@ def index
...
-    @form = current_user.forms.find_by_short_uuid(params[:id])
-    @page_num = (params[:page].present? ? params[:page].to_i : 0)
-    @page_size = (params[:page_size].present? ? params[:page_size].to_i : 500)
-    @page_size = 5000 if @page_size > 5000
+    form = current_user.forms.find_by_short_uuid(params[:id])
+    page_num = (params[:page].present? ? params[:page].to_i : 0)
+    page_size = (params[:page_size].present? ? params[:page_size].to_i : 500)
+    page_size = 5000 if page_size > 5000
+    # Date filter defaults to 1 year ago and 1 day from now
+    # Is there ever a case where we'd want to see submissions older than a year via the API?
```
ryanwoldatwork, line 16: Yes, a Form Manager should be able to see Submissions order than a year, if they specify.
I think a one year ago default works.

ryanwoldatwork (approved):

PR #541, opened 2021-07-07
TP-1129 change fixed-tab-postion to be relative on screens less than …

aktfrikshun:
TP-1129 change fixed-tab-postion to be relative on screens less than 600px wide


ryanwoldatwork:
going to close this. we can revisit later

PR #562, opened 2021-07-19
TP-1184 Hidden field support

aktfrikshun:
TP-1184 Hidden field support

### DIFF
--- a/app/views/components/forms/question_types/_hidden_field.html.erb
+++ b/app/views/components/forms/question_types/_hidden_field.html.erb
@@ -1,4 +1,4 @@
-<div class="radios" role="group" aria-labelledby="<%= question.answer_field %>">
+<div aria-labelledby="<%= question.answer_field %>">
   <%= hidden_field_tag question.answer_field.to_sym, question.placeholder_text %>
 </div>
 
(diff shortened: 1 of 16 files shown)


app/views/components/forms/question_types/_hidden_field.html.erb
```diff
@@ -1 +1,4 @@
-<%= hidden_field_tag question.answer_field.to_sym, question.text %>
+<div class="radios" role="group" aria-labelledby="<%= question.answer_field %>">
```
ryanwoldatwork, line 1: Is the div.radios necessary here?

I think the hidden field should be sufficient.

aktfrikshun: Not necessary.  I just like the way the rendered HTML flows with each field sitting in the same type of container.  But I'm cool to strip them out if desired.

ryanwoldatwork: Ah, okay to leave the div then.

but, perhaps `class` and `role` could be removed.

aktfrikshun: html display attributes removed

ryanwoldatwork (approved):

PR #695, opened 2021-10-19
TP-1309 Add versioning to websites model and controller

aktfrikshun:
TP-1309 Add versioning to websites model and controller

Adds paper trail gem, model and controller behaviors and specs


ryanwoldatwork (changes requested):
Looking good so far.

Can you add a ../websites/:id/versions path, similar to /events, showing the versions?

PR #696, opened 2021-10-25
TP-1311 Add tagging to form submissions

aktfrikshun:
TP-1311 Add tagging to form submissions

Add/Remove Tag on show page

Filter by tag on form responses page.


ryanwoldatwork (approved):
@aktfrikshun - this looks good!

I moved the tags to the sidebar, in order to group the features to process a submission together.

The left side is more just metadata. The right column is the status and now tags.

PR #698, opened 2021-11-01
TP-1324 show page with error if hidden on form submission

aktfrikshun:
TP-1324 show page with error if hidden on form submission

Update fba.js showValidationError to check if the page with the validation error is visible and, if not, show it.


ryanwoldatwork (approved):
looks good!

Please add a feature spec to test this scenario of a required question on a 2nd page.

PR #702, opened 2021-11-12
TP-1313 use partial when filtering by tag

aktfrikshun:
TP-1313 use partial when filtering by tag

### DIFF
--- a/app/controllers/admin/submissions_controller.rb
+++ b/app/controllers/admin/submissions_controller.rb
@@ -1,7 +1,7 @@
 class Admin::SubmissionsController < AdminController
   before_action :ensure_admin, only: [:feed, :export_feed]
   before_action :set_form, except: [:feed, :export_feed]
-  before_action :set_submission, except: [:feed, :export_feed]
+  before_action :set_submission, except: [:feed, :export_feed, :search]
 
   def show
   end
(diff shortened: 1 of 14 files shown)


app/controllers/admin/submissions_controller.rb
```diff
@@ -36,12 +48,12 @@ def unflag
 
   def add_tag
     @submission.tag_list.add(admin_submission_params[:tag_list].split(","))
-    @submission.save
+    @submission.save!
```
ryanwoldatwork, line 51: Why the bang here?

aktfrikshun: To raise an exception if there is a problem saving the record.   I ran into an unusual case where I had several submissions and then made a field required on the 2nd page.   When I tagged submissions I had made prior to the required field change, the tags wouldn't be saved.  No error, nothing logged.   I then discovered the submission save was not completed due to the required field validation.   I added the bang so this condition would be raised.   

If this scenario seems likely in the future, we may want to handle that case more gracefully than throwing an exception.  I can write a ticket for that if so.

aktfrikshun:
@ryanwoldatwork I added to this PR the isolating each report as a widget and loading async.  Check it out and let me know what you think.

app/controllers/admin/submissions_controller.rb
ryanwoldatwork: 👍

ryanwoldatwork (approved):

PR #708, opened 2021-12-02
TP-1331 adjust completions logic

aktfrikshun:
@ryanwoldatwork We get completion percentage over 100% when a submission is created without a corresponding activation.

Typical flow is
1) submission#new ( increments activation count )
2) submission#create ( increments response count )

The problem is that if we have a logged in user and the touchpoint is hosted then the activation is not recorded allowing for more responses than activations.   See line 13 of submissions controller new method.

My observation is that we appear to be recording a large number of form submissions without activations for hosted form "NPS.gov Was This Page Helpful" in production.   How is this possible unless a logged in user is hammering the submit button?

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -54,6 +54,14 @@ class SubmissionsController < ApplicationController
       }, status: :unprocessable_entity and return
     end
 
+    # Prevent the submission for archived forms
+    if @form && @form.archived?
+      render json: {
+        status: :unprocessable_entity,
+        messages: { submission: [t('errors.form_is_archived')] }
+      }, status: :unprocessable_entity and return
+    end
+
     @submission = Submission.new(submission_params)
     @submission.form = @form
     @submission.user_agent = request.user_agent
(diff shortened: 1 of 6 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -10,7 +10,7 @@ def new
     elsif !@form.deployable_form? && !current_user
       redirect_to index_path, alert: 'Form is not currently deployed.'
     end
-    @form.increment!(:survey_form_activations) unless current_user
+    @form.increment!(:survey_form_activations) #unless current_user
```
aktfrikshun, line 13: Always record an activation which can be followed by a submission, whether or not we have a logged in user ( keeps numbers from getting skewed )

app/models/form.rb
```diff
@@ -437,6 +437,8 @@ def omb_number_with_expiration_date
   def completion_rate
     if self.survey_form_activations == 0
       "N/A"
+    elsif (self.survey_form_activations <= self.response_count)
```
ryanwoldatwork, line 440: let's not do this.
we can address the underlying issue instead.

ryanwoldatwork (changes requested):

app/models/form.rb
aktfrikshun: Undone

app/controllers/submissions_controller.rb
aktfrikshun: I've updated this PR to only persist the submission if the form is live.   If the form is not live and we do a submit on preview we will still get a successful process flow, just no db persistence

PR #709, opened 2021-12-03
TP-1326 Optimize Submissions Table and Cache Performance.gov report

aktfrikshun:
TP-1326 Optimize Submissions Table and Cache Performance.gov report

### DIFF
--- a/app/models/form_cache.rb
+++ b/app/models/form_cache.rb
@@ -47,4 +47,9 @@ class FormCache
     Rails.cache.delete(NAMESPACE + short_uuid.to_s)
   end
 
+  def self.invalidate_reports(short_uuid)
+    Rails.cache.delete(NAMESPACE + '-performance-gov-analysis-' + short_uuid.to_s)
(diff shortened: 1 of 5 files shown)


app/models/form_cache.rb
```diff
@@ -31,6 +30,19 @@ def self.fetch_all_analysis(short_uuid)
     end
   end
 
+  def self.fetch_performance_gov_analysis(short_uuid)
```
ryanwoldatwork, line 33: I'd like to be able to force reset this cache, if needed.

When Data Collections come in, they typically come in, like 60 in a week.
I want to be sure we're able to see accurate summary data even when there's lots of activity; entering data that is used in those summary totals.

ryanwoldatwork (changes requested):

app/models/form_cache.rb
aktfrikshun: I've added a class method in FormCache to invalidate_reports.   That method will be called on the form#responses path if a parameter called use_cache is present and has a value of false.  Example:
http://lvh.me:3002/admin/forms/bd4b4302/responses?use_cache=false

PR #728, opened 2021-12-20
TP-1344 Add tags to Services and Service_Providers

aktfrikshun:
Note:  Currently there is no feature to search services and services providers.  Typically we would add a tag search under the search box.

Should we implement a search box with tags?   Should we display the tags in the table on the index pages for both services and service providers?

### DIFF
--- a/app/views/admin/service_providers/_tags.html.erb
+++ b/app/views/admin/service_providers/_tags.html.erb
@@ -1,53 +1,53 @@
 <%= form_with(model: service_provider, url: admin_service_provider_path(service_provider), local: true) do |f| %>
-	<div class="grid-row">
-	  <div class="grid-col-12">
-	  	<h3>
-				Tags
-				<%= link_to "https://github.com/gsa/touchpoints/wiki/tagging-responses", target: "_blank", rel: "noopener" do %>
-					<span class="fa fa-info-circle"></span>
-				<% end %>
-			</h3>
-	  	<p class="tag-list">
-				<ul class="usa-list usa-list--unstyled">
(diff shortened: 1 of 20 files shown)


ryanwoldatwork (approved):

app/views/admin/service_providers/_tags.html.erb
```diff
@@ -0,0 +1,53 @@
+<%= form_with(model: service_provider, url: admin_service_provider_path(service_provider), local: true) do |f| %>
```
ryanwoldatwork, line 1: @aktfrikshun - can you set your IDE to spaces? and align these tags?

app/views/admin/service_providers/index.html.erb
```diff
@@ -42,24 +42,51 @@
...
+    <%= text_field_tag :search_text, nil, class: "usa-input" %>
+    <% if @tags.size > 0 %>
+      <p>
+        Search by tag
+      </p>
+      <p class="tag-list">
+        <% @tags.order(:name).each do | tag | %>
+          <a href="#" class="search-tag-link" data-name="<%= tag.name %>">
+            <span class="usa-tag">
+              <%= tag.name %> (<%= tag.taggings_count %>)
```
ryanwoldatwork, line 57: if I added a "mobile" tag to Service Provider and Service, the label on each shows 2, when I'd expect 1.

Is there a way to scope "taggings_count" ?

app/views/admin/service_providers/_tags.html.erb
ryanwoldatwork: I updated the alignment

PR #734, opened 2021-12-26
TP-1356 Update tag_counts logic to scope to the tagged Model

aktfrikshun:
TP-1356 Update tag_counts logic to scope to the tagged Model


ryanwoldatwork (approved):
looks good.

Can you add a spec that asserts tag count is accurate to an object type?

Creating a tagged Service, then creating a tagged Website with the same tag would flesh this out.

aktfrikshun:
@ryanwoldatwork Spec added

PR #735, opened 2021-12-30
TP-1354 make tag filter link sharable

aktfrikshun:
TP-1354 make tag filter link sharable

Linked ticket: https://trello.com/c/NyNPws9g/1354-enable-tag-search-to-have-a-reliable-url

I've added the ability for the services index pages to take a "tag" parameter and filter results using that parameter.

I've also made the href of the actual tag displayed on the page to be shareable.   For example you can now right click on the tag filter and open link in new tab or copy link and share it and the selected tag filter context will be preserved.

Will this suffice for your needs?

### DIFF
--- a/app/controllers/admin/services_controller.rb
+++ b/app/controllers/admin/services_controller.rb
@@ -9,7 +9,12 @@ class Admin::ServicesController < AdminController
   ]
 
   def index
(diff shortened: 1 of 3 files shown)


ryanwoldatwork (approved):
lgtm

add a feature spec

PR #791, opened 2022-02-09
TP-1367 Add version rerporting to services, DRY up jobs and views

aktfrikshun:
TP-1367 Add version rerporting to services, DRY up jobs and views

Added vesions view and export links to Services

DRY'd up ExportVersionsJob and versions.html


ryanwoldatwork (approved):
👍

PR #823, opened 2022-03-08
Removed max-width for survey label

yosokevino:
Survey question labels were limited to max-width of 30 rem resulting in label not taking full div width. Removed max-width to allow label occupy full div width.


ryanwoldatwork:
Hi @yosokevino,

The `max-width` is part of this standard USWDS library. I am inclined to keep it.

If this form is embedded on your website, you can target the touchpoints container and the element within it, to set an alternative max-width.

PR #853, opened 2022-03-30
TP-1395 ComboBox with combo question type, states dropdown as combo, specs

aktfrikshun:
TP-1395 ComboBox implementation with new combo question type, states dropdown as combo, specs


app/views/components/widget/_widget.css.erb
```diff
@@ -62,21 +62,26 @@
   width: auto;
   z-index: 10;
 }
-// combo-box
+
+/ combo-box
```
ryanwoldatwork, line 66: //

PR #854, opened 2022-03-31
TP-1396 add label for other text input

aktfrikshun:
TP-1396 add label for other text input

This adds a label to the other text input.  

Maybe we should consider only showing this input field when the checkbox item or radio button is checked?

I don't see a use case for having more than one other option per question, regardless of locale.  Other/Other, Other/Otro, etc.  Do you @ryanwoldatwork


ryanwoldatwork:
typically, a user would use "Other" or "Otro", but not both.  

But currently, both "can" be used.

PR #865, opened 2022-04-10
TP-1411 Add webpacker and stimulus

aktfrikshun:
TP-1411 Add webpacker and stimulus

Webpacker "javascript_pack_tag" in application layout breaks some 84 test cases.  I'm guessing it's because of how we are managing javascript in a non-standard way.

I've created a separate layout for use with stimulus views only so that all test cases can pass and we can push forward with the stimulus work.


ryanwoldatwork (changes requested):
@aktfrikshun, the Snyk tests raised warnings on the webpacker versions.

`npm outdated` shows some of that can be updated.

Is it possible to update to webpack 0.572.0?
webpack-cli 4.9.2, and 
webpack-dev-server 4.8.1

PR #886, opened 2022-05-10
TP-1401 Digital Product Versions

aktfrikshun:
https://trello.com/c/RK3GPNDV/1401-enable-a-user-to-create-a-digitalproductplatform

This PR introduces the following:

- Fix CRUD for Digital Products
- Implements CRUD for Digital Product Versions
- Digital Product Versions are displayed on the "show" page of Digital Products
- Management of Digital Product Versions are contained within a Stimulus/Hotwire/TurboFrame
- DOM manipulation on delete is performed using a Stimulus/Hotwire/TurbStream response

New Design Pattern Introduced
- Decompose pages into components using TurboFrames
- Manipulate component DOM using event triggered TurboStreams


spec/views/admin/digital_product_versions/create.html.erb_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
```
ryanwoldatwork, line 1: we haven't been using these view specs,
so these can be rm'd

ryanwoldatwork (approved):

PR #909, opened 2022-06-03
USDR relations

ryanwoldatwork:
enables a Registry user to specify many Users, as contacts,
and specify many Organizations, as agencies
for both Digital Product and Digital Service Account

### DIFF
--- a/app/controllers/admin/digital_products_controller.rb
+++ b/app/controllers/admin/digital_products_controller.rb
@@ -139,6 +139,20 @@ class Admin::DigitalProductsController < AdminController
     end
   end
 
+  def search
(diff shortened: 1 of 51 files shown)


app/controllers/admin/digital_products_controller.rb
```diff
@@ -57,6 +60,38 @@ def remove_tag
     @digital_product.save
   end
 
+  def add_organization
+    @organization = Organization.find_by_id(params[:organization][:id])
+
+    if @organization
+      binding.pry
```
aktfrikshun, line 67: Probably shouldn't be left here ;-)

PR #911, opened 2022-06-06
TP-1427 Add persona select to service stage

aktfrikshun:
https://trello.com/c/CaA6dZRa/1427-service-manager-can-reference-a-persona-in-a-service-stage

Add belongs_to, has_many relationships to user, persona
Add migration to add persona_id to service_stages
Seed 3 personas for admin user
Add persona select to service stage create/update form
Add persona to service stage show page
Add persona factory to specs
Add creation of 3 personas to user factory
Update service stage feature spec to choose and verify presence of selected persona


app/models/user.rb
```diff
@@ -13,6 +13,7 @@ class User < ApplicationRecord
   has_many :user_roles, dependent: :destroy
   has_many :forms, through: :user_roles, primary_key: "form_id"
   has_many :collections, through: :organization
+  has_many :personas
```
ryanwoldatwork, line 16: the User won't be attached to a Persona. so, this can be removed.

More-formally assigning/identifying persona-type roles is an interesting idea. (Could be great for formalized-workflows)
But, we don't have a requirement to set, or for users to set their own Persona.

The Persona objects will be co-managed in the CX process for the first round of service definitions (happening this year). Agency Service Managers will be able to associate stages with Personas.

db/seeds.rb
```diff
@@ -64,6 +64,21 @@ def production_suitable_seeds
 admin_user.save!
 puts 'Created Admin User: #{admin_user.email}'
 
+Persona.create!({
+  name: 'Public User',
+  user_id: admin_user.id
```
ryanwoldatwork, line 69: no `user_id` relation

app/models/user.rb
aktfrikshun: I was curious about you set user_id as required for persona.   Maybe we can discuss at tomorrows standup how these persona's are going to be used?   If user_id is required for a persona then it seems like the has_many/belongs_to between users and persona should be explicit.

ryanwoldatwork: oh, the `user_id` was for indicating the user who created the Persona. this will likely change in a similar way to Websites, where multiple people can manage one.

the "ownership" or "governance" around the Personas are still evolving.

aktfrikshun: Ok,  I removed the relationship between persona and user.

PR #915, opened 2022-06-09
TP-1426-website-manager-can-assign-many-personas-to-a-website

aktfrikshun:
https://trello.com/c/mV9HfBYG/1426-website-manager-can-assign-many-personas-to-a-website

Ryan, what do you think about using tags to associate Personas with Websites ( and other many to many's )?

- Add Persona to Website: @website.persona_list.add(persona_id)
- Remove Persona from Website: @website.persona_list.remove(persona_id)
- Get Website personas: @website.persona_list
- All websites with Persona:  Website.tagged_with(persona_id, :on => :personas)
- 
Screen Shot 2022-06-09 at 11 15 56 AM


ryanwoldatwork:
using tags for many to many relationships makes sense!

app/views/admin/websites/_website_personas.html.erb
```diff
@@ -0,0 +1,42 @@
...
+            <span class="fa fa-trash"></span>
+          </a>
+        </li>
+          <% end %>
+      </ul>
+      <p>
+    </div>
+  </div>
+  <%= select_tag :website_persona_id, options_for_select(@website_persona_options.map { |p| [p.name, p.id] }), prompt: "Choose a Persona", class: "usa-select add-website-persona" %>
+<script>
```
ryanwoldatwork, line 23: wrap the query in a `$(function() {
  // the jquery code goes here. otherwise, it may not load (due to browser race conditions)
})`

PR #933, opened 2022-06-16
TP-1444 Run rubocop autocorrect on Layout category issues only

aktfrikshun:
TP-1444 Run rubocop autocorrect on Layout category issues only

This will make indentation, whitespace, etc consistent across the whole project.


app/controllers/admin/collections_controller.rb
```diff
@@ -1,9 +1,8 @@
 class Admin::CollectionsController < AdminController
   before_action :set_collection, only: [:show, :edit, :copy,
-    :submit, :publish,
-    :update, :destroy,
-    :events
-  ]
+                                        :submit, :publish,
```
ryanwoldatwork, line 3: @aktfrikshun - is there any way to not have this amount of indentation?

app/controllers/admin/collections_controller.rb
```diff
@@ -12,30 +11,30 @@ def index
     if performance_manager_permissions?
       if @quarter && @year
         @collections = Collection.where(quarter: @quarter, year: @year)
-          .order('organizations.name', :year, :quarter, 'service_providers.name')
-          .includes(:organization, :service_provider)
+                                 .order('organizations.name', :year, :quarter, 'service_providers.name')
```
ryanwoldatwork, line 14: I'd prefer to keep a single tab indent here, instead of what's happening here.

PR #939, opened 2022-06-21
TP-1450 Ensure search works for digital products

aktfrikshun:
https://trello.com/c/9DGvJmra/1450-ensuring-the-search-filters-work-on-digital-service-accounts

- Fixed org, service, keyword searches
- Added aasm filter
- Added specs

@ryanwoldatwork  What do you think about adding search by sponsoring agencies here?   Or is that more for reporting?


ryanwoldatwork (changes requested):
looks good! 

can you remove the `debugger`?

ryanwoldatwork (approved):

PR #967, opened 2022-07-08
Website actions

ryanwoldatwork:


app/controllers/admin/websites_controller.rb
```diff
@@ -228,16 +249,77 @@ def approve
...
+
+    def reset
+      if @website.reset!
+        Event.log_event(Event.names[:website_reset], 'Website', @website.id, "Website #{@website.domain} was reset at #{DateTime.now}", current_user.id)
+        redirect_to admin_website_url(@website), notice: "Website #{@website.domain} has been reset to newly_requested status."
+      else
+        render :edit
+      end
+    end
+
```
aktfrikshun, line 322: Can we DRY this up a bit?

I am thinking we can define the valid transitions [:develop, :stage, :launch, etc] and define a single controller method to transition state.

app/views/admin/websites/tasks/_bar.html.erb
```diff
@@ -0,0 +1,97 @@
...
+      <div class="grid-col-12">
+        <%= link_to reset_admin_website_path(website), class: "usa-button usa-button--secondary full-width", method: :post do %>
+        <span class="fa fa-backward"></span>
+        &nbsp;
+        Reset
+        <% end %>
+      </div>
+    </div>
+  <% end %>
+</div>
```
aktfrikshun, line 97: Same thing here...

Can we DRY up the view such that we simply have a single component which takes a block for the state transition condition and returns a links to the next possible states?

config/routes.rb
```diff
@@ -170,6 +170,13 @@
         get 'collection_request', to: 'websites@collection_request', as: :collection_request
         post 'approve', to: 'websites#approve', as: :approve
         post 'deny', to: 'websites#deny', as: :deny
+        post 'develop', to: 'websites#develop', as: :develop
+        post 'stage', to: 'websites#stage', as: :stage
+        post 'launch', to: 'websites#launch', as: :launch
+        post 'redirect', to: 'websites#redirect', as: :redirect
+        post 'archive', to: 'websites#archive', as: :archive
+        post 'decommission', to: 'websites#decommission', as: :decommission
+        post 'reset', to: 'websites#reset', as: :reset
```
aktfrikshun, line 179: Replace these with transition_state?

aktfrikshun (approved):
Just a couple suggestions for DRY'ing things up for your consideration

app/controllers/admin/websites_controller.rb
ryanwoldatwork: ya, refactoring this makes sense!

PR #1270, opened 2023-10-19
Show required fields label only if form includes required fields

aduth:
We have a simple Yes/No optional questionnaire on our website ([example](https://login.gov/help/trouble-signing-in/overview/)), and I noticed recently the addition of a "required field" label since #1231. The label feels rather out of place for our simple form, which doesn't include any required fields.

The changes proposed here would only display this label on a form which includes required fields.


ryanwoldatwork:
awesome. thanks @aduth!

PR #1425, opened 2024-05-23
Introduce USWDS theming and upgrade to USWDS 3.8.

sanason:
This PR does 2 things:

1. Updates the version of USWDS we are using from 3.6.1 to 3.8. This will give us access to a new sticky table header feature which we can use for https://trello.com/c/7f3i1ylK/79-when-a-table-is-longer-than-viewport-table-header-row-stays-visible-on-scroll.
2. Adopts a somewhat more formal procedure for upgrading and customizing USWDS in our project (documented in the README).

Number 2 is just a proposal, from a brand new member of the team who has no context on why things are done the way they are done so, reviewers, feel free to criticize. That said, I think the new procedure offers these benefits:

- Clearer documentation of which USWDS version we are on (since it's now recorded in `package.json`).
- Single automated command that ensures that all updated assets are copied to the right place (`gulp updateUswds`).
- Follows the "approved" approach for customizing USWDS (by adjusting the settings provided by the framework) which makes our customizations less likely to break when USWDS is updated. (Note that this PR only takes one tiny step in this direction.)

#### Did the upgrade to 3.8 break anything in the Touchpoints app? Did the upgrade to 3.8 break anything in the display of questionnaire forms?
No changes that I can see. The only breaking change listed (in [3.8](https://github.com/uswds/uswds/releases/tag/v3.8.0)) concerns the class `usa-layout-docs__sidenav` which, as far as I can tell, we aren't really using.
#### 

#### Do our tables look the same after using the `theme-table-border-color` setting to customize their appearance?
Yes, borders look the same.


app/assets/stylesheets/site.scss
```diff
@@ -251,21 +251,6 @@ tbody tr:hover td {
   text-decoration: none;
 }
 
-.usa-table {
```
sanason, line 254: These overrides were replaced by setting `$theme-table-border-color` in `uswds/_uswds-theme.scss`.

app/assets/stylesheets/application.css
```diff
@@ -10,7 +10,7 @@
  * files in this directory. Styles in this file should be added after the last require_* statement.
  * It is generally better to create a new file per style scope.
  *
- *= require uswds
```
sanason, line 13: Renamed because I couldn't figure out a way to get the USWDS compiler to generate the file with its original name.

app/assets/stylesheets/uswds-styles.css
sanason, line 1: I wish this was not minified but that's how the USWDS compiler operates. There is a feature request in [uswds-compile](https://github.com/uswds/uswds-compile/issues/80) asking to make minification optional. Admittedly, that is a very old feature request and the project hasn't had a release for a long time but I could ask about it.

sanason:
> Follows the "approved" approach for customizing USWDS (by adjusting the settings provided by the framework) which makes our customizations less likely to break when USWDS is updated. 

Here's an example of customizations being broken:
Originally, I upgraded to 3.8 simply by copying over the new versions of `uswds.js` and `uswds.css`. Then I added the new class `usa-table--sticky-header` to the list-of-forms table and got this result:
Screenshot 2024-05-23 at 3 24 40 PM
The new class added a border to the table and, since we weren't setting `theme-table-border-color`, we ended up with a black border.

ryanwoldatwork (approved):
@sanason - thank you!

this is looking good and accomplishes:

* a better uswds build process
* minified versions of the assets are in place
* updated assets (icons)
* kept style customizations
* removed old, non-minified assets

PR #1436, opened 2024-05-31
Combobox

ryanwoldatwork:


sanason:
Screenshot 2024-06-04 at 10 54 59 AM

Maybe the "Other" option is confusing here since it suggests that the user will be able to enter their own "other" text but the combobox on the generated form doesn't actually support that.

sanason (approved):
No comments other than the one about the "Other" option.

PR #1444, opened 2024-06-07
don't reset the form after it is submitted

ryanwoldatwork:
currently, when a form is submitted, the "Success" message shows,
and when a form's modal is closed, the form resets to a state as if to receive another response, which is not an intention.


sanason (approved):
I'm not sure I fully understand the customer ask from the Trello ticket:

> Can we hide the survey from to users who've taken it
Possibly, using GTM
Can a form be triggered to show after a specific amount of time on a site?
Can a form be triggered to show after a specific amount of clicks on a site?

That said, the changes in this PR make sense to me.

Issue #1446, opened 2024-06-18
Support CSV download for large response sets


sanason:
Support download through the web UI for at least several 100K responses. 

Rough justification for how to choose a max size to support: If the exported file is small enough to import directly into Google sheets (or Excel, maybe), then we should be able to export that file. Once the file is bigger than that, the user is going to need more sophisticated tools to deal with their data, so  we can probably get away with asking them to use the API to retrieve their response data.

Google sheets limits:

> Up to 10 million cells or 18,278 columns for spreadsheets imported from Microsoft Excel. The limits are the same for Excel and CSV imports.

Since our CSV export has 13 non-question columns, that means ~700K responses for a single-question survey and ~300K responses for a survey with the maximum number of questions (20).

ryanwoldatwork:
Given current response volume, supporting 200-300k downloads would cover MOST clients. 
More responses than that can use the API.

I've been instructing users with more than 10,000 responses to use the API.

PR #1447, opened 2024-06-18
Improve client-side handling of failed websocket downloads.

sanason:
The main change here is in what happens if the websocket gets disconnected after the download has been requested but before the response has been received.

Previously, the code would try to reconnect the websocket and, once connected, it would re-request the data. The second download would probably fail too, and the third, and so forth. During all of this, the UI would give no indication of what was happening - it would just continue to display the "Exporting..." label on the export button.

Now, the code gives up after the first disconnect. It then resets the export button and, in place of the data file it was supposed to return, it returns a file with the following text:

> We're sorry, your download has failed. That happens sometimes for response sets over several MB's in size.
>
> We're working on fixing this problem. You can monitor our progress at `https://github.com/GSA/touchpoints/issues/1446`. In the meantime, you can use our API to download large response sets. View API documentation at `https://github.com/GSA/touchpoints/wiki/API`.

This was the easiest way to signal an error but it's probably not the best user experience. I'm happy to consider a different approach. I'm also happy to change the text.

The PR also fixes some minor problems where the code didn't always restore the original text to the export buttons once exports were complete (i.e., the button read as "Export FY2022" on page load, transitioned to "Exporting..." during export and then transitioned to "Export FY Responses".)

### DIFF
--- a/app/assets/javascripts/channels/export.js
+++ b/app/assets/javascripts/channels/export.js
@@ -1,15 +1,19 @@
-function subscribeExportChannel(uuid, startExportJobFtn, completionCallback) {
+function subscribeExportChannel(uuid, url, completionCallback) {
   App.cable.subscriptions.create(
     { channel: "ExportChannel", uuid: uuid },
     {
       connected: function() {
-        startExportJobFtn();
+        $.get(url);
       },
 
       // Called when the WebSocket connection is closed, either by server or by client-side stale connection monitor.
       // If the connection closes before we receive a response, count that as a failed export and unsubscribe.
       // Reconnecting is not useful because we can't tell whether the response was sent while we were disconnected.
       disconnected: function() {
+        if (newrelic) {
+          newrelic.noticeError(new Error("Export channel disconnected."), { requestedUrl: url});
+        }
+
         const errorText = "We're sorry, your download has failed. That happens sometimes for response sets over several MB's in size.\n\n" +
           "We're working on fixing this problem. You can monitor our progress at https://github.com/GSA/touchpoints/issues/1446. " +
           "In the meantime, you can use our API to download large response sets. View API documentation at https://github.com/GSA/touchpoints/wiki/API.";
--- a/app/assets/javascripts/channels/export.js
+++ b/app/assets/javascripts/channels/export.js
@@ -1,37 +1,42 @@
-function subscribeExportChannel(uuid, callback) {
-  App.download = App.cable.subscriptions.create(
+function subscribeExportChannel(uuid, startExportJobFtn, completionCallback) {
+  App.cable.subscriptions.create(
     { channel: "ExportChannel", uuid: uuid },
     {
       connected: function() {
-        callback();
+        startExportJobFtn();
       },
 
-      disconnected: function() {},
+      // Called when the WebSocket connection is closed, either by server or by client-side stale connection monitor.
+      // If the connection closes before we receive a response, count that as a failed export and unsubscribe.
+      // Reconnecting is not useful because we can't tell whether the response was sent while we were disconnected.
+      disconnected: function() {
+        const errorText = "We're sorry, your download has failed. That happens sometimes for response sets over several MB's in size.\n\n" +
+          "We're working on fixing this problem. You can monitor our progress at https://github.com/GSA/touchpoints/issues/1446. " +
+          "In the meantime, you can use our API to download large response sets. View API documentation at https://github.com/GSA/touchpoints/wiki/API.";
+
+        const blob = new Blob([errorText], {
+          type: "text/plain;charset=utf-8"
+        });
+
+        saveAs(blob, `touchpoints-failed-export-${(new Date().toISOString().slice(0, 19))}`);
+
+        // 'this' is the subscription object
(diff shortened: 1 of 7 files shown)


config/routes.rb
```diff
@@ -418,8 +418,8 @@
     get 'performance/apg', to: 'performance#apgs', as: :apgs
     get 'management', to: 'site#management', as: :management
     get 'events', to: 'events#index', as: :events
-    get 'events/:id', to: 'events#show', as: :event
```
sanason, line 421: While testing all of the code that uses `subscribeExportChannel`, I discovered that `GET /admin/events/export` was broken because it was incorrectly matching `events/:id`. Reordering the routes fixed this.

app/assets/javascripts/channels/export.js
```diff
@@ -1,37 +1,42 @@
...
-          .html("Export A11v2 Responses to CSV")
-          .removeClass('cursor-not-allowed');
+        const blob = new Blob([data.csv], {
+          type: "text/csv;charset=utf-8"
+        });
+
+        saveAs(blob, data.filename);
 
-        App.download.unsubscribe();
-        App.cable.disconnect();
```
sanason, line 32: We want this subscriber to unsubscribe once their download is complete but I don't think we should go further and disconnect the websocket - there could be other downloads in progress. We might end up holding open an unused websocket for a while but I think that's okay. It will be closed once the user navigates to a different page.

config/environments/development.rb
```diff
@@ -49,7 +49,7 @@
   # Store uploaded files on the local file system (see config/storage.yml for options).
   config.active_storage.service = :local
 
-  config.active_job.queue_adapter = :sidekiq
```
sanason, line 52: Ryan, I don't know how you have your development environment set up. Is this change okay?

ryanwoldatwork: I do typically use sidekiq locally.

But I'm okay with commenting this out with an additional comment like `used for testing Sidekiq locally`

app/assets/javascripts/channels/export.js
```diff
@@ -1,37 +1,42 @@
...
-        callback();
+        startExportJobFtn();
       },
 
-      disconnected: function() {},
+      // Called when the WebSocket connection is closed, either by server or by client-side stale connection monitor.
+      // If the connection closes before we receive a response, count that as a failed export and unsubscribe.
+      // Reconnecting is not useful because we can't tell whether the response was sent while we were disconnected.
+      disconnected: function() {
+        const errorText = "We're sorry, your download has failed. That happens sometimes for response sets over several MB's in size.\n\n" +
```
sanason, line 13: I'd like to add instrumentation here, so we can keep track of how often this failure happens. What's the best way to do that? New Relic PageAction event?

app/assets/javascripts/channels/export.js
```diff
@@ -1,37 +1,42 @@
-function subscribeExportChannel(uuid, callback) {
-  App.download = App.cable.subscriptions.create(
+function subscribeExportChannel(uuid, startExportJobFtn, completionCallback) {
```
ryanwoldatwork, line 1: maybe `startingCallback` ?

just thinking about consistency with `completionCallback`

ryanwoldatwork (changes requested):
looks great!

can you ensure the other `subscribeExportChannel` are addressed?

config/environments/development.rb
sanason: > I do typically use sidekiq locally.

In that case, I'll revert this change. I can get used to running sidekiq locally.

app/assets/javascripts/channels/export.js
sanason: Yeah, that's a better suggestion. Implemented.

app/views/admin/events/index.html.erb
```diff
@@ -41,18 +41,23 @@
...
+      const uuid = generateUUID();
+      const buttonHtml = $(this).html();
       $(this).html('Exporting...').addClass('cursor-not-allowed');
 
-      var url = $(this).attr('href') + '?uuid=' + uuid + '';
-      subscribeExportChannel(uuid, function() {
-        $.get(url);
-      });
+      const url = $(this).attr('href') + '?uuid=' + uuid + '';
+      subscribeExportChannel(
```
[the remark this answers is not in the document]
sanason: I counted 6 total when I searched for `subscribeExportChannel`. Do you think there are more?

app/assets/javascripts/channels/export.js
sanason: I switched to passing in the `url` since I needed it for error reporting.

app/views/admin/events/index.html.erb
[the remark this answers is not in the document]
sanason: And I updated those 6 and tested them manually.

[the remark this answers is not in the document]
ryanwoldatwork: i thought that's what a Find yielded..   yet.   thanks for addressing them!

ryanwoldatwork (approved):
nice updates! 👍

PR #1472, opened 2024-07-12
Performance improvements for responses page

sanason:
This PR does 2 things:

1. The first commit improves the loading performance of the "Responses by Status" table on the form responses page. Login.gov mentioned that it wasn't loading for their giant "Login.gov yes/no survey", now it loads in less than 1 second.
2. The second commit provides support for exporting large response sets to csv. The old implementation tried to pass the exported csv file back over a websocket (which often failed when the file was too big), the new implementation saves the exported csv file to S3 and passes a link to the file back over a websocket. I've tested this locally and got the following results:

- On our 3rd largest response set (NPSgov CX, 35K responses), the export succeeds in ~20 seconds.
- For exports over 300K responses, the export request fails immediately. Tested successfully with our 2nd largest response set (Login.gov yes/no survey, over 2M responses).
- For the 'Login.gov yes/no survey', the FY2022 response set is less than 300K (I don't have the exact number, ~150K responses). This response set downloaded successfully in ~2 minutes.

We need additional S3 configuration to go along with this PR:

- (Required) [Set a CORS policy](https://cloud.gov/docs/services/s3/#allowing-client-side-web-access-from-external-applications) on our private bucket
- (Recommended) [Set a lifecycle configuration policy](https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-expire-general-considerations.html) to expire objects with the `temporary_files` prefix in our private bucket.

### DIFF
--- a/app/assets/javascripts/channels/export.js
+++ b/app/assets/javascripts/channels/export.js
@@ -3,26 +3,23 @@ function subscribeExportChannel(uuid, url, completionCallback) {
     { channel: "ExportChannel", uuid: uuid },
     {
       connected: function() {
-        $.get(url);
+        $.ajax(url)
+          .fail(response => {
+            this.unsubscribe();
+            if (response.status === 400 && response.responseText) {
+              this.downloadError(response.responseText);
+            } else {
+              this.downloadError();
+            }
+            completionCallback();
+          });
       },
 
       // Called when the WebSocket connection is closed, either by server or by client-side stale connection monitor.
       // If the connection closes before we receive a response, count that as a failed export and unsubscribe.
       // Reconnecting is not useful because we can't tell whether the response was sent while we were disconnected.
       disconnected: function() {
-        if (newrelic) {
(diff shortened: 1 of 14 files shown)


app/assets/javascripts/channels/export.js
```diff
@@ -3,44 +3,55 @@ function subscribeExportChannel(uuid, url, completionCallback) {
...
       received: function(data) {
-
-        const blob = new Blob([data.csv], {
-          type: "text/csv;charset=utf-8"
-        });
-
-        saveAs(blob, data.filename);
+        const dataUrl = data.url;
+        const filename = data.filename;
+        $.ajax(dataUrl)
```
sanason, line 32: This call requires that the bucket be configured for CORS.

ryanwoldatwork (approved):
thanks @sanason 

this improves how the app handles larger exports!

app/assets/javascripts/channels/export.js
ryanwoldatwork: i think we'll move to the async email soon

PR #1483, opened 2024-07-18
Add arm64 platform to Gemfile.lock.

sanason:
Solves a problem I was having installing `nokogiri` on my Apple M1 Pro.


ryanwoldatwork (approved):
👍

PR #1486, opened 2024-07-22
Load USWDS modal js and css with the Touchpoints widget

sanason:
Should not change any user-facing behavior.

Two changes to our generated third-party script:

1. The script adds a deferred `<script>` tag to the customer's HTML, which downloads a static `touchpoints-uswds` script from Touchpoints. This script contains USWDS JavaScript, just the JS from the `modal` package. The script is bundled as an immediately-invoked function expression so it should not contaminate the customer's global JS scope. Once the DOM is built, the script adds the USWDS modal event handlers to just those parts of the DOM that we constructed, so if the customer is also using some version of the USWDS modal, the event handlers that we add will not fire on any part of the customer's DOM.
2. The script adds a `<link>` tag to the customer's HTML, which downloads a static `touchpoints-widget.css` stylesheet from Touchpoints.

JS/CSS bundling options:

1. Perform manually and commit the result.
3. Enable Rails css-bundling and js-bundling.


app/assets/stylesheets/touchpoints-widget.css
```diff
@@ -0,0 +1,15 @@
+/*
+ * This is a manifest file that'll be compiled into application.css, which will include all the files
```
ryanwoldatwork, line 2: is it possible to include this CSS in the `widget.css.erb` file?

Combining the CSS would keep the same number of calls from a client's browser to Touchpoints, without adding an extra call for .css in this case.

app/views/components/widget/_fba.js.erb
```diff
@@ -82,12 +86,28 @@ function FBAform(d, N) {
 			});
 
 		},
+		addModalJs() {
+			const scriptElement = document.createElement('script');
+			scriptElement.type = 'text/javascript';
+			scriptElement.defer = true;
+			scriptElement.src = "<%= asset_url('touchpoints-uswds.min.js') %>"
```
ryanwoldatwork, line 93: one consideration here is to see if this asset gets hashed differently each deploy.

if so, I'm wondering about a scenario where a client copies this code to host on their own, but the asset reference goes stale between deployments.

gulpfile.js
```diff
@@ -12,9 +23,51 @@ uswds.paths.dist.fonts = './public/fonts';
...
-exports.updateUswds = uswds.updateUswds;
\ No newline at end of file
+exports.updateUswds = uswds.updateUswds;
+
+// Tasks to build USWDS assets for the embedded survey widget.
+// These tasks do not use the USWDS compiler.
+exports.bundleWidgetJS = bundleWidgetJS;
+exports.compileWidgetSass = compileWidgetSass;
+exports.buildWidgetAssets = parallel(bundleWidgetJS, compileWidgetSass);
+
```
sanason, line 40: These gulp tasks are largely copied from https://github.com/uswds/uswds/tree/develop/tasks

PR #1494, opened 2024-07-24
Load USWDS modal JS and CSS.

sanason:
> [!NOTE]  
> This PR should not change any user-facing behavior.

This PR does 3 things:

**Number 1**
It introduces two new asset files, a JavaScript file and a CSS file. Both are built from USWDS code and both are meant to be used with the survey widget, not with the Touchpoints app. The JS file contains JS code for the USWDS modal (and its dependencies), bundled with Browserify. The CSS file contains CSS code for the USWDS modal (and its dependencies), compiled by SASS. These assets will be included in the survey widget whenever:
1. The widget is embedded (not hosted on Touchpoints) AND
2. The owner of the survey has indicated that they want Touchpoints to provide its own USWDS code instead of inheriting USWDS from the base site.

**Number 2**
`_fba.js.erb` and `_fba2.js.erb` have been updated to load these 2 new assets when appropriate. This should have no effect on the user-visible behavior of the widget because the CSS classes used in the new assets don't match the CSS classes used in the widget. A second PR will update the CSS classes used in the widget and thus switch to using the USWDS modal in the widget.

**Number 3**
The rest of this PR adds infrastructure to support building the 2 new assets from USWDS code.


app/assets/stylesheets/widget-uswds-styles.css
sanason, line 1: Since the base site may have loaded its own copy of USWDS code and may have written its own CSS rules involving `usa-` classes, here we add a `tp-` prefix to our uses of the `usa-` classes in order to avoid conflicts.

gulpfile.js
```diff
@@ -12,9 +27,78 @@ uswds.paths.dist.fonts = './public/fonts';
...
+    .pipe(dest(uswds.paths.dist.js))
+    .pipe(sourcemaps.init({ loadMaps: true }))
+    .pipe(uglify())
+    .pipe(rename({ suffix: ".min" }))
+    .pipe(sourcemaps.write("."))
+    .pipe(dest(uswds.paths.dist.js));
+}
+
+// TODO: Add browserslistrc
+async function compileWidgetSass() {
```
sanason, line 71: Copied with minor changes from the [USWDS repository](https://github.com/uswds/uswds/blob/develop/tasks/sass.js). The main change is that I added a `prefixer` step which adds the `tp-` prefix to our version of the USWDS classes.

app/views/components/widget/_fba2.js.erb
```diff
@@ -82,6 +85,21 @@ function FBAform(d, N) {
 				d.head.appendChild(style);
 			}
 		},
+		_loadUswds: function() {
```
sanason, line 88: I'm running up against my Rail inexperience here and I'm open to other ways of loading these 2 static files for the widget, including ways that maintain the single downloadable javascript payload. Requesting these assets separately seems reasonable to me - I don't think it will have a noticeable impact on performance, assuming we minify and cache the assets, maybe even put them in a CDN one day. The only users who might notice are those with `load_css=true` who are self-hosting a copy of their widget's javascript.

app/assets/stylesheets/widget-uswds-styles.css
```diff
@@ -0,0 +1,163 @@
...
+}
+.tp-usa-js-no-click .tp-usa-modal,
+.tp-usa-js-no-click .tp-usa-modal *{
+  pointer-events:auto;
+  -webkit-user-select:text;
+     -moz-user-select:text;
+          user-select:text;
+}
+
+[data-open-modal] *{
```
ryanwoldatwork, line 17: @sanason - did this somehow escape the namespace?

ryanwoldatwork (changes requested):
this looks good.

i made a few changes in a branch that was branched from this branch...
with the goal to inline the css and .js generated in this PR, rather than load them in separate http requests

https://github.com/GSA/touchpoints/pull/1504/commits

gulpfile.js
```diff
@@ -12,9 +27,78 @@ uswds.paths.dist.fonts = './public/fonts';
...
\ No newline at end of file
+exports.updateUswds = uswds.updateUswds;
+
+// Tasks to build USWDS assets for the embedded survey widget.
+// These tasks do not use the USWDS compiler.
+exports.bundleWidgetJS = bundleWidgetJS;
+exports.compileWidgetSass = compileWidgetSass;
+exports.buildWidgetAssets = parallel(bundleWidgetJS, compileWidgetSass);
+
+// TODO: Do we need to babelify?
```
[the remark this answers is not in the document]
sanason: So that the USWDS JS is built with support for older browsers - I believe that's the benefit.

app/views/components/widget/_fba.js.erb
```diff
@@ -93,6 +96,21 @@ function FBAform(d, N) {
 			d.head.appendChild(style);
 		},
 		<% end %>
+		loadUswds: function() {
+			const cssTag = document.createElement('link');
+			cssTag.rel = 'stylesheet';
+			cssTag.type = 'text/css';
+			cssTag.media = 'all';
+			cssTag.href = "<%= asset_url('widget-uswds-styles.css') %>";
```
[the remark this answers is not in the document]
sanason: No longer using asset urls.

app/views/components/widget/_fba.js.erb
```diff
@@ -93,6 +96,21 @@ function FBAform(d, N) {
...
+			document.getElementsByTagName('head')[0].appendChild(cssTag);
+
+			const scriptElement = document.createElement('script');
+			scriptElement.type = 'text/javascript';
+			scriptElement.defer = true;
+			scriptElement.src = "<%= asset_url('touchpoints-uswds.js') %>"
```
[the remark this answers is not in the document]
sanason: Everything rendering in single script now.

app/views/components/widget/_fba2.js.erb
```diff
@@ -658,6 +676,9 @@ const touchpointFormOptions<%= form.short_uuid %> = {
 	'elementSelector': "<%= form.element_selector %>",
 	'css' : "<%= escape_javascript(render partial: 'components/widget/widget', formats: :css, locals: { form: form }) %>",
 	'loadCSS' : <%= form.load_css %>,
+	'loadUswds' : <%= form.load_css && form.delivery_method != "touchpoints-hosted-only" %>,
+	'uswdsCssUrl' : "<%= asset_url('widget-uswds-styles.css') %>",
```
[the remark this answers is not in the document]
sanason: New css is now rendered within app/views/components/widget/_widget.css.erb.

ryanwoldatwork (approved):

app/assets/stylesheets/widget-uswds-styles.css
sanason: The prefixer doesn't prefix attributes. Would a prefixed `data` attribute be valid HTML?

This seems to be the only escapee. I'm going to comment it out for the time being and revisit when I switch our code to use the USWDS modal.

PR #1513, opened 2024-07-30
Add babelify transform to the bundleWidgetJS task.

sanason:
Adds a `babelify` step to the `bundleWidgetJS` task, which runs against the list of browsers supported by USWDS.


.browserslistrc
sanason, line 1: Copied from [USWDS repo](https://github.com/uswds/uswds/blob/369cf3fdf9abf721a357e864a0f6adbf667ba862/.browserslistrc).

For comparison, `defaults` translates to `> 0.5%, last 2 versions, Firefox ESR, not dead`.

app/views/components/widget/_widget-uswds.js.erb
sanason, line 1: As far as I can tell, running Babel didn't actually add any polyfills, so maybe it's unnecessary for the modal package given USWDS's supported browser policy. But I'm leaning toward adding the babelify step anyway since the [USWDS bundling task](https://github.com/uswds/uswds/blob/develop/tasks/javascript.js) includes it.

babel.config.js
sanason, line 1: I'm deleting this because the new `babelify` task was using this file for configuration, which was not what I wanted.

I believe this file is an artifact from Rails 6, which used webpacker, and isn't needed anymore, now that we're on Rails 7. I could be wrong about that though.

postcss.config.js
sanason, line 1: This doesn't need to be deleted for this PR to work but I do think that is also a Rails 6 artifact (like babel.config.js) and is no longer needed.

ryanwoldatwork (approved):

PR #1522, opened 2024-08-02
Use the USWDS modal in the embedded widget.

sanason:
Use the USWDS modal in the embedded widget. It looks like this:

Screenshot 2024-08-07 at 6 06 05 PM

### CSS class name namespacing

Our traditional method for "namespacing" our CSS rules (scoping our rules inside our own container class) doesn't work for the modal.

Let's say this is the modal DOM code that our `fba2.js` script adds to the base website, with the idea that `.fba` will be our namespacing class:

```html
 <div class="fba">
    <div class="usa-modal fba-modal" id="fba-modal-a09c8a7c" aria-labelledby="fba-form-title-a09c8a7c">
        <div class="usa-modal__content fba-modal-dialog">
            <div class=usa-modal__main>
                <div class="touchpoints-form-wrapper">
                    <div class="wrapper">
                        <div class="word-break fba-modal-title">
                            <h2 id="fba-form-title-a09c8a7c">Product feedback form</h2>
                        </div>
                        ...form contents
                    </div>
                </div>
            </div>
            <button class="usa-modal__close fba-modal-close" data-close-modal>X</button>
        </div>
    </div>
 </div>   
<button data-open-modal aria-controls="fba-modal-a09c8a7c">Click to open</button>
```

When the USWDS modal JS code runs, it reconstructs the modal portion of the DOM to look like this (all modals are moved to the end of the DOM, which "allows us to more easily find the elements to hide from screen readers when the modal is open"):

```html
<div class="fba">
  <div data-placeholder-for="fba-modal-a09c8a7c" aria-hidden="true" data-original-class="usa-modal fba-modal" data-original-id="fba-modal-a09c8a7c" data-original-aria-labelledby="fba-form-title-a09c8a7c" style="display: none;"></div>
</div>
<button data-open-modal aria-controls="fba-modal-a09c8a7c">Click to open</button>
<div class="usa-modal-wrapper is-hidden" role="dialog" id="fba-modal-a09c8a7c" aria-labelledby="fba-form-title-a09c8a7c" data-opener="fba-button">
  <div class="usa-modal-overlay" aria-controls="fba-modal-a09c8a7c">
    <div class="usa-modal fba-modal" data-touchpoints-form-id="a09c8a7c" tabindex="-1">
      <div class="usa-modal__content fba-modal-dialog" role="dialog" aria-modal="true">
        <div class="usa-modal__main">
          <div class="touchpoints-form-wrapper " id="touchpoints-form-a09c8a7c" data-touchpoints-form-id="a09c8a7c" data-focus="">
            <div class="wrapper">
              <div class="word-break fba-modal-title">
                <h2 id="fba-form-title-a09c8a7c">Was this page helpful?</h2>
              </div>
              ...
            </div>
          </div>
          <button class="fba-usa-modal__close fba-modal-close" type="button" aria-label="Close this window" data-close-modal="">×</button>
        </div>
      </div>
    </div>
  </div>

```

Looking at the reconstructed HTML, certain classes like `usa-modal-wrapper` and `usa-modal-overlay` cannot be namespaced by wrapping them in `.fba` because they will be constructed outside our container elements. I.e., this doesn't work:

```
.fba .usa-modal-wrapper.is-hidden{
  visibility:hidden;
  opacity:0;
  position:fixed;
}
```

They _can_ be namespaced with prefixes because the USWDS JS code can be configured to add prefixes. So, this works:

```
.fba-usa-modal-wrapper.is-hidden{
  visibility:hidden;
  opacity:0;
  position:fixed;
}
```

Or we could just not namespace these CSS rules at all:

```
.usa-modal-wrapper.is-hidden{
  visibility:hidden;
  opacity:0;
  position:fixed;
}
```

but then there's a danger that we're altering the appearance of the base website, which might have its own `usa-modal` elements which have received custom styling.


app/views/components/widget/_fba.js.erb
```diff
@@ -13,7 +13,6 @@ function FBAform(d, N) {
 		formElement: function() {
 			return this.formComponent().querySelector("form");
 		},
-		activatedButton: null, // tracks a reference to the button that was clicked to open the modal
```
ryanwoldatwork, line 16: fwiw, i'm okay with treating this file as legacy (going to be deprecated)

app/views/components/widget/_fba2.js.erb
```diff
@@ -100,15 +89,14 @@ function FBAform(d, N) {
 		<%# inject the form interface for modal and custom-button-modal %>
 		if (this.options.deliveryMethod && (this.options.deliveryMethod === 'modal' || this.options.deliveryMethod === 'custom-button-modal')) {
 			this.dialogEl = d.createElement('div');
-			this.dialogEl.setAttribute("hidden", true);
-			this.dialogEl.setAttribute("aria-hidden", true);
-			this.dialogEl.setAttribute('class', 'fba-modal');
+			this.dialogEl.setAttribute('class', "<%= form.prefix "usa-modal" %> fba-modal");
+			this.dialogEl.setAttribute('id', this.modalId());
+			this.dialogEl.setAttribute('aria-labelledby', `fba-form-title-${this.options.formId}`);
+			this.dialogEl.setAttribute('aria-describedby', 'something');
```
sanason, line 95: The modal code requires that you set `aria-describedby` to something non-empty. I'm not sure what to set it to in our case.

app/views/components/widget/_fba2.js.erb
```diff
@@ -416,22 +386,6 @@ function FBAform(d, N) {
 				d.body.prepend(this.skipLink);
 			}
 		},
-		// Used when in a modal
-		loadDialog: function() {
-			d.dispatchEvent(new Event('onTouchpointsModalOpen'));
```
sanason, line 421: I haven't figured out how to retain the `onTouchpointsModalOpen` and `onTouchpointsModalClose` events. How important are these events?

app/views/components/widget/_modal.html.erb
```diff
@@ -2,12 +2,19 @@
...
   role="dialog"
-  aria-label="Feedback modal dialog"
-  aria-labelledby="fba-modal-title"
   aria-modal="<%= ['modal', 'custom-button-modal'].include?(form.delivery_method) ? 'true' : 'false' %>">
+  <div class="<%= form.prefix "usa-modal__main" %>">
   <%= render "components/widget/no_modal", form: form do %>
     <%= render partial: 'components/forms/footer', locals: { form: form } %>
   <% end %>
+  </div>
+  <%- unless form.delivery_method == "inline" %>
```
sanason, line 13: USWDS accessibility guidance says:

> Include the “X” close button at the end of the modal code.

gulpfile.js
```diff
@@ -65,9 +64,6 @@ async function bundleWidgetJS() {
 async function compileWidgetSass() {
   const pluginsProcess = [
     discardComments(),
-    customPrefix({
```
sanason, line 68: This is still being done but it's being done in SASS so that it's easier to apply it to some styles but not to other styles.

gulpfile.js
```diff
@@ -53,6 +51,7 @@ async function bundleWidgetJS() {
       aliases: {
         '../../uswds-core/src/js/config': './uswds/uswds-config.js'
       },
+      appliesTo: { regex: '.*/usa-modal/.*'},
```
sanason, line 54: Only the modal JS code is getting prefixed.

uswds/widget-uswds-styles/_prefixed-usa-modal.scss
```diff
@@ -0,0 +1,181 @@
+@use "uswds-core" as * with (
+  $theme-modal-default-max-width: "tablet"
+);
```
sanason, line 3: This changes the max width of the modal from 30rem to 40rem. Our previous setting was 35rem. The max width of the form element is 30rem and the modal looks weird if its max width is not bigger than the form's max width. I also tried the large version of the modal but it has too much padding, too much white space, IMO.

app/views/components/widget/_widget.css.erb
```diff
@@ -1033,11 +972,6 @@
   text-decoration:none;
 }
 
-/* Override */
-.fba-modal-dialog .usa-form {
-  max-width: inherit;
-}
```
sanason, line 1039: This rule was making the input fields too wide. They looked like this:

Screenshot 2024-08-05 at 3 17 02 PM

app/views/components/widget/_fba2.js.erb
ryanwoldatwork: the events should remain, if possible

https://github.com/GSA/touchpoints/wiki/Analytics-Events

app/views/components/forms/_footer_banner.html.erb
```diff
@@ -9,7 +9,7 @@
             alt="U.S. flag"
           />
       </div>
-      <div class="grid-col-fill tablet:grid-col-auto" aria-hidden="true">
```
sanason, line 12: Fixes row 2 and row 7 in vote.gov's accessibility issues.

uswds/widget-uswds-styles/_prefixed-usa-modal.scss
sanason, line 1: This file is the one USWDS file that's not being handled by the build. I copied it and manually added some prefixes. I had a way to add prefixes to all of our imported USWDS SCSS files but I haven't found a nice way to add prefixes to just some of the files. That is going to have to wait for later work.

app/views/components/widget/_fba2.js.erb
sanason: I restored `onTouchpointsModalOpen`, the easy one.

spec/rails_helper.rb
```diff
@@ -78,7 +78,7 @@
     # with the specs, so continue to use transaction strategy for speed.
     driver_shares_db_connection_with_specs = Capybara.current_driver == :rack_test
 
-    page.driver.browser.manage.window.resize_to(1200, 800)
+    page.driver.browser.manage.window.resize_to(1280, 1024)
```
sanason, line 81: Changed this to fix a failure in spec/features/embedded_touchpoints_spec.rb. I think the modal content is narrower and thus longer now so some of the DOM elements referenced in the test were not visible without scrolling. To avoid adding scrolling statements in the test and to avoid redesigning the form in the test, I made the window bigger. I concede that this can't be the solution every time we change the size of some element but 1280x1024 is a pretty reasonable window size. On the other hand, if we specifically chose the original size for the tests, I can put it back.

app/views/components/widget/_fba2.js.erb
sanason: I will create a ticket to try to restore `onTouchpointsModalClose`.

PR #1542, opened 2024-08-15
Disable USWDS-based version of embedded modal for all forms.

sanason:
Start the roll-out of the USWDS embedded modal by setting all forms to _not_ use the USWDS-based version, until I've done some testing of the new code on existing forms in production.

### DIFF
--- a/db/migrate/20240815183449_enable_legacy_form_embed_for_all_forms.rb
+++ b/db/migrate/20240815183449_enable_legacy_form_embed_for_all_forms.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+# Feature flag migration to accompany release of USWDS-based modal
+class EnableLegacyFormEmbedForAllForms < ActiveRecord::Migration[7.1]
+  def up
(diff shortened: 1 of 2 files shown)


db/migrate/20240815183449_enable_legacy_form_embed_for_all_forms.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# Feature flag migration to accompany release of USWDS-based modal
+class EnableLegacyFormEmbedForAllForms < ActiveRecord::Migration[7.1]
+  def up
+    execute 'UPDATE forms SET legacy_form_embed = TRUE'
+  end
+
+  def down
```
sanason, line 9: This is a legitimate revert because I just verified that `legacy_form_embed` is false for all forms in production.

ryanwoldatwork: 👍

ryanwoldatwork (approved):

PR #1549, opened 2024-08-16
Compile widget after382

ryanwoldatwork:


sanason (commented):
Should we also update the USWDS assets used by the Touchpoints app (run `npx gulp updateUswds`)?

ryanwoldatwork:
ya, if there's a conventional way to run the tasks

sanason:
`npx gulp updateUswds` updates the assets for the Touchpoints app.

https://github.com/GSA/touchpoints/wiki/Developer-Guide#updating-uswds

sanason (changes requested):
I think it'll cause confusion if we don't update both the app and widget when we upgrade the USWDS version.

https://github.com/GSA/touchpoints/wiki/Developer-Guide#updating-uswds

ryanwoldatwork:
I added a npm script `updateUswds` to run all the commands.

Doing so yielded the following diff.
But it looks like that can all be ignored. Thoughts?

fsddsfsdasfdfsa

sanason (approved):
👍

sanason:
> Doing so yielded the following diff. But it looks like that can all be ignored. Thoughts?

Yes, agreed. In particular, I have a [USWDS PR](https://github.com/uswds/uswds/pull/6026) out to prevent the changes in `_widget-uswds.js.erb`.

PR #1552, opened 2024-08-20
org does not allow special chars for an abbr

ryanwoldatwork:


sanason (approved):
I made one suggestion but looks good.

PR #1558, opened 2024-08-22
Send form inactivity emails

ryanwoldatwork:


.circleci/cron.sh
```diff
@@ -23,6 +23,7 @@ cf run-task touchpoints-staging-sidekiq-worker -c "rake scheduled_jobs:send_dail
 cf run-task touchpoints-staging-sidekiq-worker -c "rake scheduled_jobs:send_weekly_notifications"
 cf run-task touchpoints-staging-sidekiq-worker -c "rake scheduled_jobs:check_expiring_forms"
 cf run-task touchpoints-staging-sidekiq-worker -c "rake scheduled_jobs:archive_forms"
+cf run-task touchpoints-staging-sidekiq-worker -c "rake scheduled_jobs:notify_form_managers_of_inactive_forms"
```
ryanwoldatwork, line 26: runs job in staging and demo,
but not prod here

app/models/form.rb
```diff
@@ -41,6 +41,18 @@ def target_for_delivery_method
     errors.add(:element_selector, "can't be blank for an inline form") if (delivery_method == 'custom-button-modal' || delivery_method == 'inline') && (element_selector == '')
   end
 
+  def roles
+    user_roles.map { |role| { role: role.role, user: role.user }}
+  end
+
+  def form_managers
```
ryanwoldatwork, line 48: could select from user_roles here too...  but tried to keep the change minimal (moving this from the view)

app/models/form.rb
```diff
@@ -242,6 +254,22 @@ def set_uuid
...
+
+    inactive_published_forms.each do |form|
+      user_emails = form.form_managers.reject { |user| user.inactive }.collect(&:email)
+      UserMailer.form_inactivity_email(form_short_uuid: form.short_uuid, user_emails: user_emails, days_ago: days_ago).deliver_later
+      puts "sending for #{form.short_uuid} to #{user_emails}"
+    end
+  end
+
+  def self.find_inactive_forms_since(days_ago)
+    min_time = Time.now - days_ago.days
```
ryanwoldatwork, line 268: only forms X days old, so we're not messaging days 31, 32, 33 too

app/views/admin/forms/show.html.erb
```diff
@@ -65,16 +65,13 @@
             <%= label_tag "Organization", nil, class: "usa-label margin-bottom-1" %>
             <%= render "admin/organizations/badge", organization: @form.organization %>
           </div>
-          <% @roles = @form.user_roles.map { |role| { role: role.role, user: role.user }} %>
```
ryanwoldatwork, line 68: moved to form

spec/mailers/user_mailer_spec.rb
```diff
@@ -17,8 +17,8 @@
     end
 
     it 'renders the body' do
-      expect(mail.body.encoded).to match('Touchpoints.gov Response Notification')
```
ryanwoldatwork, line 20: updated all the `match` to `have_text` which is a more conventional assertion

MaroyaF (approved):

PR #1561, opened 2024-08-23
Fixes issue where embedded touchpoints form interferes with existing USWDS elements on a page.

sanason:
USWDS reports that some of their preview widgets aren't working (on https://designsystem.digital.gov/components/combo-box/, for example) on pages that also have a Touchpoints modal on them. They suggest the problem is that the Touchpoints code is calling init on USWDS components within the base USWDS page, instead of limiting its scope to Touchpoints elements only. This PR should fix that problem and prevent Touchpoints from messing with the page it is embedded on.

I think it's possible that, with this code, 2 Touchpoints forms on the same page could interfere with and break each other. But that seems like an edge case, to be addressed later.

### DIFF
--- a/app/views/components/widget/_widget-uswds.js.erb
+++ b/app/views/components/widget/_widget-uswds.js.erb
@@ -3907,8 +3907,11 @@ const DatePicker = require("@uswds/uswds/packages/usa-date-picker/src/index.js")
 const Modal = require("@uswds/uswds/packages/usa-modal/src/index.js");
 
 // Initialize event listeners
-ComboBox.on();
-DatePicker.on();
-Modal.on();
+const fbaFormElement = document.querySelector(".touchpoints-form-wrapper");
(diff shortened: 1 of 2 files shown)


app/views/components/widget/_widget-uswds.js.erb
```diff
@@ -3907,8 +3907,11 @@ const DatePicker = require("@uswds/uswds/packages/usa-date-picker/src/index.js")
 const Modal = require("@uswds/uswds/packages/usa-modal/src/index.js");
 
 // Initialize event listeners
-ComboBox.on();
-DatePicker.on();
-Modal.on();
+const fbaFormElement = document.querySelector(".touchpoints-form-wrapper");
+if (fbaFormElement) {
```
ryanwoldatwork, line 3911: 👍

ryanwoldatwork (approved):

PR #1562, opened 2024-08-23
fix cx volume count

ryanwoldatwork:


app/models/cx_collection_detail.rb
```diff
@@ -10,7 +10,7 @@ class CxCollectionDetail < ApplicationRecord
   validates :channel, presence: true
 
   def volume_of_respondents
-    CxResponse.count
```
ryanwoldatwork, line 13: this column is returning the total number of all CxResponses, rather than the CxResponses for only this CxCollectionDetail.

this updates, that, and adds a test to ensure each row of the .csv is showing its response count, rather than the sum

MaroyaF (approved):

PR #1568, opened 2024-08-26
remove service_id from CxCollectionDetail and use the CxCollection.service instead

ryanwoldatwork:
* removes the CxCollectionDetail.service_id field, because its CxCollection has a service_id field and it should use that
* this required model relationships changes regarding how the CxCollection accesses a Service and vice versa
* the `service_id` is ultimately retained for reporting purposes; CxCollectionDetail just looks it up through its CxCollection

### DIFF
--- a/app/controllers/admin/cx_collection_details_controller.rb
+++ b/app/controllers/admin/cx_collection_details_controller.rb
@@ -31,7 +31,6 @@ class Admin::CxCollectionDetailsController < AdminController
 
   def create
     @cx_collection_detail = CxCollectionDetail.new(cx_collection_detail_params)
-    @cx_collection_detail.service_id = @cx_collection_detail.cx_collection.service_id
 
     respond_to do |format|
       if @cx_collection_detail.save
@@ -154,6 +153,6 @@ class Admin::CxCollectionDetailsController < AdminController
     end
 
     def cx_collection_detail_params
-      params.require(:cx_collection_detail).permit(:cx_collection_id, :service_id, :transaction_point, :channel, :service_stage_id, :volume_of_customers, :volume_of_customers_provided_survey_opportunity, :volume_of_respondents, :omb_control_number, :federal_register_url, :reflection_text, :survey_type, :survey_title, :trust_question_text)
+      params.require(:cx_collection_detail).permit(:cx_collection_id, :transaction_point, :channel, :service_stage_id, :volume_of_customers, :volume_of_customers_provided_survey_opportunity, :volume_of_respondents, :omb_control_number, :federal_register_url, :reflection_text, :survey_type, :survey_title, :trust_question_text)
     end
 end
--- a/app/controllers/admin/services_controller.rb
+++ b/app/controllers/admin/services_controller.rb
@@ -84,7 +84,10 @@ module Admin
       send_data @services.to_csv, filename: "touchpoints-services-#{Date.today}.csv"
     end
 
-    def show; end
+    def show
+      @omb_cx_reporting_collections = @service.omb_cx_reporting_collections.includes(:collection).order("collections.year", "collections.quarter")
+      @cx_collections = @service.cx_collections
+    end
 
     def new
       @service = Service.new
--- a/app/models/cx_collection_detail.rb
+++ b/app/models/cx_collection_detail.rb
@@ -1,5 +1,5 @@
 class CxCollectionDetail < ApplicationRecord
-  belongs_to :service
+  belongs_to :service, through: :cx_collection
   belongs_to :cx_collection
   belongs_to :service_stage, optional: true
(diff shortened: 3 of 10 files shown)


app/controllers/admin/cx_collection_details_controller.rb
```diff
@@ -31,7 +31,6 @@ def upload
 
   def create
     @cx_collection_detail = CxCollectionDetail.new(cx_collection_detail_params)
-    @cx_collection_detail.service_id = @cx_collection_detail.cx_collection.service_id
```
ryanwoldatwork, line 34: no longer need to set the service_id here (the CxCollection already has it)

app/controllers/admin/services_controller.rb
```diff
@@ -84,7 +84,10 @@ def export_csv
       send_data @services.to_csv, filename: "touchpoints-services-#{Date.today}.csv"
     end
 
-    def show; end
+    def show
```
ryanwoldatwork, line 87: these are moved from the view to here

app/models/cx_collection_detail.rb
```diff
@@ -1,10 +1,10 @@
 class CxCollectionDetail < ApplicationRecord
-  belongs_to :service
   belongs_to :cx_collection
   belongs_to :service_stage, optional: true
   has_many :cx_responses, dependent: :delete_all
   has_one :cx_collection_detail
   has_one :service_provider, through: :cx_collection
+  has_one :service, through: :cx_collection
```
ryanwoldatwork, line 7: was belongs_to
now has_one

app/models/service.rb
```diff
@@ -14,8 +14,8 @@ class Service < ApplicationRecord
   has_many :omb_cx_reporting_collections
   has_many :collections, through: :omb_cx_reporting_collections
 
-  has_many :cx_collection_details
-  has_many :cx_collections, through: :cx_collection_details
+  has_many :cx_collections
```
ryanwoldatwork, line 17: reversed this lookup

app/serializers/cx_collection_detail_serializer.rb
```diff
@@ -22,15 +22,19 @@ def service_provider_id
       object.service_provider.id if object.service_provider
     end
 
+    def service_provider_name
+      object.service_provider.name if object.service_provider
+    end
+
+    def service_id
```
ryanwoldatwork, line 29: keep this for reporting

app/views/admin/services/show.html.erb
```diff
@@ -288,9 +288,6 @@
   </div>
 </div>
 
-<% @omb_cx_reporting_collections = @service.omb_cx_reporting_collections.includes(:collection).order("collections.year", "collections.quarter") %>
```
ryanwoldatwork, line 291: moved to the controller

app/views/admin/services/show.html.erb
```diff
@@ -299,10 +296,18 @@
   <table class="usa-table">
     <thead>
       <tr>
-        <th>Fiscal year</th>
-        <th>Quarter</th>
-        <th>Collection name</th>
-        <th>Service name</th>
+        <th data-sortable scope="col">
```
ryanwoldatwork, line 299: make these sortable

app/views/admin/services/show.html.erb
```diff
@@ -331,30 +336,38 @@
...
+        </th>
+        <th data-sortable scope="col">
+          Service name
+        </th>
       </tr>
     </thead>
     <tbody>
-      <% @cx_collection_details.each do |cx_collection_detail| %>
-      <tr>
+      <% @cx_collections.each do |cx_collection| %>
```
ryanwoldatwork, line 358: render the collections instead of the details

docs/models/cx_collections.md
```diff
@@ -17,7 +17,6 @@ Data is published for download at https://www.performance.gov/cx/data/.
 ++| organization_name (Service provider    | integer  | Organization Name                                                           | General Services Administration       |
 | quarter                                  | string   | The quarter of the fiscal year                                              | 2                                     |
 | rating                                   | string   | The rating for the service                                                  | true,false,partial                    |
-| service_id                               | integer  | Identifier for the service                                                  | 74                                    |
```
sanason, line 20: I'm surprised this is being removed from `cx_collections.md` rather than from `cx_collection_details.md`

sanason (approved):
I have one question about the docs but it's not a blocker.

sanason:
Thanks for all the comments!

docs/models/cx_collections.md
ryanwoldatwork: good catch here. the docs should be cx collection details indeed.
will update.

PR #1574, opened 2024-08-27
Set feedback-analytics as the return-path for our emails

sanason:
This should cause bounce and complaint notifications to be sent to feedback-analytics. Following the instructions in [Amazon SES Email feedback forwarding destination](https://docs.aws.amazon.com/ses/latest/dg/monitor-sending-activity-using-notifications-email.html#monitor-sending-activity-using-notifications-email-destination). We already have email feedback forwarding enabled, we just need to update where the notifications are sent.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -76,7 +76,8 @@ Rails.application.configure do
 
   config.active_job.queue_adapter = :sidekiq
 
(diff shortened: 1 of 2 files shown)


ryanwoldatwork (approved):
https://api.rubyonrails.org/classes/ActionMailer/Base.html

sanason:
@ryanwoldatwork , thanks for the quick approval. I can't really test this in dev.

PR #1581, opened 2024-08-28
support login.gov all_emails

ryanwoldatwork:
* rather than the `email`, which is the last used email (which could be a personal address)

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -4,7 +4,7 @@ module Users
   class OmniauthCallbacksController < Devise::OmniauthCallbacksController
     def login_dot_gov
       @kind = 'Login.gov'
-      @email = auth_hash['info']['email'] if auth_hash && auth_hash['info']['email_verified']
+      @email = auth_hash['info']['all_emails'].find { |email| email.end_with(".gov") || email.end_with(".mil") } if auth_hash && auth_hash['info']['email_verified']
 
       login
     end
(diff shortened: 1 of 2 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -4,7 +4,7 @@ module Users
   class OmniauthCallbacksController < Devise::OmniauthCallbacksController
     def login_dot_gov
       @kind = 'Login.gov'
-      @email = auth_hash['info']['email'] if auth_hash && auth_hash['info']['email_verified']
+      @email = auth_hash['info']['all_emails'].find { |email| email.end_with(".gov") || email.end_with(".mil") } if auth_hash && auth_hash['info']['email_verified']
```
ryanwoldatwork, line 7: select the first gov or mil email, rather than using the `email` which is the last email used to login

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -30,7 +30,7 @@ def auth_hash
     def login
       Event.log_event(Event.names[:user_authentication_attempt], 'Event::Generic', 1, "Email #{@email} attempted to authenticate on #{Date.today}")
 
-      @user = User.from_omniauth(auth_hash) if @email.present?
+      @user = User.from_omniauth(auth_hash, @email) if @email.present?
```
ryanwoldatwork, line 33: pass the selected email explicity

app/models/user.rb
```diff
@@ -49,10 +49,10 @@ def self.admins
     User.where(admin: true)
   end
 
-  def self.from_omniauth(auth)
+  def self.from_omniauth(auth, email)
     # Set login_dot_gov as Provider for legacy TP Devise accounts
     # TODO: Remove once all accounts are migrated/have `provider` and `uid` set
-    @existing_user = User.find_by_email(auth.info.email)
+    @existing_user = User.find_by_email(email)
```
ryanwoldatwork, line 55: find the user by the passed in, selected email, and not the last used email

and below, set the specified email

MaroyaF (approved):

PR #1586, opened 2024-08-30
Prefix the 'visible' class in embedded widget.

sanason:
A couple of our users' modal surveys began behaving strangely when they were switched to the new USWDS-based modal. In its hidden state, the modal was blocking content on the underlying page. Affected surveys were at:

- https://catalog.data.gov/dataset/series-information-for-2020-census-urban-growth-area-uga-state-based-tiger-line-shapefiles-curr
- https://www.faa.gov/regulations_policies/airworthiness_directives

Both sites are using Bootstrap, which defines a visibility CSS rule:

```
.visible {
  visibility: visible !important;
}
```

This is a case where the CSS on the base site interferes with the styling of our embedded widget. We also use a `visible` class on an element in the modal:

```
<div class="section visible">
```

The USWDS modal conceals itself when closed by applying `visibility: hidden; opacity: 0` to itself.

Put this all together and the `section` element inside our survey is styled as:

.touchpoints-form-wrapper form div.section.visible {
    display: block;
}

.visible {
    visibility: visible !important;
}

.fba-usa-modal-wrapper.is-hidden {
   ~visibility: hidden;~
    opacity: 0;
    position: fixed;
}

which means that [it is invisible but it still accepts pointer events](https://stackoverflow.com/a/34529598).

### DIFF
--- a/spec/factories/form.rb
+++ b/spec/factories/form.rb
@@ -198,6 +198,7 @@ FactoryBot.define do
     trait :kitchen_sink do
       name { 'Kitchen Sink Form 🧼' }
       kind { 'custom' }
+      title { 'Kitchen Sink Form' }
       after(:create) do |f, _evaluator|
         FactoryBot.create(:question,
                           form: f,
@@ -228,7 +229,7 @@ FactoryBot.define do
                           position: 20,
                           text: "Some custom <a href='#'>html</a>")
 
-        option_elements_section = f.form_sections.create(title: 'Option elements', position: 2)
+        option_elements_section = f.form_sections.create(title: 'Page 2', position: 2)
         radio_button_question = FactoryBot.create(:question,
                                                   form: f,
                                                   form_section: option_elements_section,
@@ -286,9 +287,11 @@ FactoryBot.define do
                                  position: 3,
                                })
 
+        custom_elements_section = f.form_sections.create(title: 'Page 3', position: 3)
+
         dropdown_question = Question.create!({
                                                form: f,
-                                               form_section: option_elements_section,
+                                               form_section: custom_elements_section,
                                                text: 'Custom Question Dropdown',
                                                question_type: 'dropdown',
                                                help_text: 'This is help text for a dropdown.',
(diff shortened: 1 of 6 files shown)


spec/features/embedded_touchpoints_spec.rb
```diff
@@ -23,9 +23,13 @@
         context 'default success text' do
           before do
             click_on('Help improve this site') # opens modal
-            expect(page).to have_content('Help improve this site')
 
+            expect(page).to have_content('Help improve this site')
             expect(page).to have_content('Do you have a few minutes to help us test this site?')
+
+            expect(page).to have_content('Page 1')
```
sanason, line 30: Test that page 1 is visible and page 2 and 3 are not.

ryanwoldatwork: also consider `expect(page).to_not have_content('Page 2')`

ryanwoldatwork (approved):

spec/features/embedded_touchpoints_spec.rb
sanason: Changed.

spec/factories/form.rb
sanason, line 1: I made some changes to the kitchen sink form so that it would fit on the page without scrolling, since content that needs to be scrolled to seems to cause flakey tests.

ryanwoldatwork: ah.. okay... yes, i've seen those flakey specs before

but i'd like to try to keep the form intact if possible.

i'll take a look at the specs and see if the failure looks familiar

sanason: It's very nearly the same form. I shortened the title, removed 1 option from one multi-option question and moved one question to another page. I think the test is logically the same.

ryanwoldatwork: given this test changes visibility, the specs are saying something here.
it may not be material.
but capybara is peculiar about selection, and i'd hope to get it working. i saw a similar issue last week and was able resolve it.

PR #1587, opened 2024-08-31
Sort forms by status

ryanwoldatwork:


app/views/admin/forms/index.html.erb
sanason, line 1: Is there a way to eliminate the `commit=Filter` param?

Screenshot 2024-09-03 at 12 17 44 PM

ryanwoldatwork: 👍
yep, by removing the name attribute. coming up

PR #1593, opened 2024-09-03
spec looks like race condition

ryanwoldatwork:
* .. when creating records from csv
* ensure cx responses are loaded


spec/controllers/api/v1/cx_responses_controller_spec.rb
```diff
@@ -57,6 +57,8 @@
         end
 
         it 'return a default paginated array of cx_responses' do
+          expect(cx_collection_detail.cx_responses.count).to eq(1000)
```
ryanwoldatwork, line 60: ensures these 1000 records exist first.

the 1000 records are loaded in this text as cx_responses, from a CSV fixture (seed file)

MaroyaF (approved):

PR #1594, opened 2024-09-03
introduce spam score

ryanwoldatwork:
* TBD on ways to calculate it and convey it to a user

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_08_26_211558) do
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -816,6 +816,7 @@
     t.string "aasm_state", default: "received"
     t.string "hostname"
     t.string "tags", default: [], array: true
+    t.integer "spam_score", default: 0
```
sanason, line 819: How do we know integer will be the right type for this?

sanason (approved):

db/schema.rb
ryanwoldatwork: a demo i saw returned an integer.

we don't.
but its a place to start and can be changed.

PR #1605, opened 2024-09-05
Don't rely on order of S3 instances in VCAP_SERVICES.

sanason:
Map the S3 instances in VCAP_SERVICES by plan type instead of relying on the order in which the instances are listed, to guarantee that private objects are going in the private bucket.


ryanwoldatwork (approved):
👍

PR #1614, opened 2024-09-09
test pages for js errors

ryanwoldatwork:


app/views/admin/goal_targets/edit.html.erb
```diff
@@ -3,11 +3,13 @@
 <%= link_to 'Cancel', admin_goal_path(@goal), class: "usa-button cancel-goal-target" %>
 
 <script>
+$(function() {
```
sanason, line 6: I assume that adding these wrapping `function` calls fixed some JS error. What was the error?

ryanwoldatwork: This is basically DOM.ready, in jquery $(function() {}).  The test errors prompted me to see this.  This change is a nice-to-have here.

The error: `Uncaught ReferenceError: $ is not defined` was caused by the /goal_targets page being loaded directly in a test, when, in the app, it is a subpage of /goals.  Loading the page directly renders without a layout, and thus the jquery error.

sanason (approved):

PR #1623, opened 2024-09-10
Fix bug where tabbing in modal gets stuck on close button.

sanason:
This is a workaround for [this USWDS bug](https://github.com/uswds/uswds/issues/6056).

### DIFF
--- a/app/views/components/forms/_custom.html.erb
+++ b/app/views/components/forms/_custom.html.erb
@@ -5,17 +5,6 @@
   class="usa-form usa-form--large margin-bottom-2"
   method="POST">
   <div class="touchpoints-form-body">
-    <%= hidden_field_tag :authenticity_token, form_authenticity_token if form.verify_csrf? -%>
-    <%= hidden_field_tag :fba_location_code, params[:location_code], tabindex: -1 %>
-    <input
-      type="text"
-      name="fba_directive"
-      id="fba_directive"
-      class="display-none"
-      title="fba_directive"
-      aria-hidden="true"
-      tabindex="-1"
-      autocomplete="off">
     <% form.form_sections.each_with_index do |section, index| %>
       <div class="section <%= 'fba-visible' if section == form.form_sections.first -%>">
         <%- if (section != form.form_sections.first) || (section != form.form_sections.last) %>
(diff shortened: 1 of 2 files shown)


spec/rails_helper.rb
```diff
@@ -52,7 +52,7 @@
 
 RSpec.configure do |config|
   # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
-  config.fixture_paths = "#{::Rails.root}/spec/fixtures"
```
sanason, line 55: Rubocop fix.

spec/rails_helper.rb
```diff
@@ -41,9 +41,9 @@
 Capybara.register_driver :selenium do |app|
   Capybara::Selenium::Driver.new(app, browser: :chrome)
 end
-Capybara.javascript_driver = :selenium                 # Run feature specs with Firefox
+# Capybara.javascript_driver = :selenium                 # Run feature specs with Firefox
 # Capybara.javascript_driver = :selenium_chrome          # Run feature specs with Chrome
-# Capybara.javascript_driver = :selenium_chrome_headless # Run feature specs with headless Chrome
+Capybara.javascript_driver = :selenium_chrome_headless # Run feature specs with headless Chrome
```
ryanwoldatwork, line 46: what do these changes get us? 
are they required?

spec/rails_helper.rb
ryanwoldatwork: 👍

app/views/components/forms/_custom.html.erb
```diff
@@ -127,5 +116,16 @@
         <% end %>
       </div>
     <% end %>
+    <%= hidden_field_tag :authenticity_token, form_authenticity_token if form.verify_csrf? -%>
```
ryanwoldatwork, line 119: looks like these hidden fields were moved.

how were they impacting the tabbing?
is it only ordering?

spec/rails_helper.rb
sanason: Sorry, should have commented on this change.

This reverses a change you made in a [recent PR](https://github.com/GSA/touchpoints/pull/1613/files#diff-053aa3ea46cb7c6a649e0d0fc592ce729b66e1d44be475516fd58857ba695862). I was guessing that you changed from headless to headed browser for debugging purposes and didn't mean to commit it. I noticed the change because now, when I run rspec locally, the headed browser being used for testing basically takes over the input devices on my computer and I can't type or use the keyboard.

app/views/components/forms/_custom.html.erb
sanason: Yes, only the ordering changes. The linked USWDS ticket tries to explain why hidden inputs are affecting the tabbing. We can avoid tabbing issues as long as the hidden inputs are listed somewhere in the middle of the modal code, neither as the first element or the last element.

ryanwoldatwork (approved):

app/views/components/forms/_custom.html.erb
ryanwoldatwork: 👍

spec/rails_helper.rb
ryanwoldatwork: 🤦‍♂️

👍

PR #1627, opened 2024-09-12
remove legacy_form_embed

ryanwoldatwork:
date: 2024-09-12

note: this branch will probably hang around til early October before we can ensure all users are migrated off the legacy form markup


app/controllers/admin/forms_controller.rb
```diff
@@ -237,11 +237,7 @@ def example
     end
 
     def js
-      if @form.legacy_form_embed
-        render(partial: 'components/widget/fba', formats: :js, locals: { form: @form })
-      else
-        render(partial: 'components/widget/fba2', formats: :js, locals: { form: @form })
-      end
+      render(partial: 'components/widget/fba2', formats: :js, locals: { form: @form })
```
sanason, line 240: Suggest renaming the remaining file to `_fba`. fba2 without fba is confusing.

sanason (approved):
One suggestion, otherwise looks good. I'm glad to be done with this!

app/controllers/admin/forms_controller.rb
ryanwoldatwork: good thought!

will do

PR #1631, opened 2024-09-18
Set max width and border on 'new' version of inline widget.

sanason:
The code changes that switched the modal version of our widget to use the USWDS modal also changed the appearance of the inline version of the widget. This PR restores the original appearance of the inline widget.

Inline widget before the USWDS modal changes:
Screenshot 2024-09-18 at 10 18 34 AM

Inline widget after the USWDS modal changes:
Screenshot 2024-09-18 at 10 19 32 AM

Inline widget after this PR:

Screenshot 2024-09-18 at 10 21 24 AM

### DIFF
--- a/app/views/components/widget/_widget.css.erb
+++ b/app/views/components/widget/_widget.css.erb
@@ -98,6 +98,11 @@
     max-width: inherit;
 }
 <% else %>
+.fba-modal-dialog {
+  font-family: Source Sans Pro Web, Helvetica Neue, Helvetica, Roboto, Arial, sans-serif;
+  font-size: 100%;
+}
+
 .fba-modal-dialog h1 {
   margin-right: 20px;
   margin-top: -1rem;
--- a/app/views/components/widget/_widget.css.erb
+++ b/app/views/components/widget/_widget.css.erb
@@ -105,6 +105,13 @@
 }
 <% end %>
 
+.fba-inline-container {
+  background: #fff;
+  border: 1px solid #E5E5E5;
+  max-width: 35rem;
+  margin: 0 auto 40px auto;
+}
+
 .fixed-tab-button {
   bottom: 0;
   padding: 5px 10px;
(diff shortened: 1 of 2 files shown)


app/views/components/widget/_widget.css.erb
```diff
@@ -105,6 +105,13 @@
 }
 <% end %>
 
+.fba-inline-container {
+  background: #fff;
+  border: 1px solid #E5E5E5;
```
ryanwoldatwork, line 110: i'd be inclined to exclude the border for now.

the width is and centering of the element (not necessarily the text) is what's important

As an example, see https://demo.touchpoints.digital.gov's `#touchpoints-yes-no-form` example, where a few customizations have been added.

sanason: This PR shouldn't affect inline widgets with `suppress_submit_button == true`, which the `#touchpoints-yes-no-form` form is an example of, I think. It only affects the kind of inline widgets that have a submit button and the Touchpoints banner at the bottom.

ryanwoldatwork: `suppress_submit_button` should only apply to a single question form with yes/no buttons.  The option is about attaching the form 'submit' event to each yes and no button, and not requiring an additional submit button.

I'd expect these styles to apply to all inline forms, even with that flag.
Specifically, we'd want the yes/no buttons to be subject to a limited width, centered form component too.

ryanwoldatwork (approved):

app/views/components/widget/_widget.css.erb
sanason: This PR is intended to get inline forms to look like they did before the introduction of the USWDS modal. Inline forms that previously were centered and bordered and max-width'ed will be so again. The yes/no button forms , which have never been centered and bordered (I think they do have a max-width), will continue to not be centered and bordered.

I'd like to match the original styling and get everyone off the original form and _then_ think about changing/improving the styling.

ryanwoldatwork (approved):

PR #1644, opened 2024-09-30
Add an admin page to manage record retention and deletion.

sanason:
This PR is meant to address https://trello.com/c/ClAzKNNP (run report on unique form owners) but goes a little beyond that.

I added an admin page to manage record retention and deletion.

Screenshot 2024-09-30 at 4 51 22 PM

On this page, you can view a list of forms scheduled for deletion and export the list of emails of affected form managers.

Screenshot 2024-09-30 at 4 52 44 PM

Tomorrow, the list will be much longer, since today is the last day of the fiscal year.

I tried to take the text of the relevant records schedule at its word:

> Cutoff at the end of the fiscal year in which the posting becomes superseded, obsolete, or canceled. Destroy 3 years after cutoff. 

When is a form superseded, obsolete or cancelled? "When it is archived" seems the most intuitive answer to me. I propose we start auto-archiving forms that haven't received a response in a long time. Then, if someone actually goes in and resets the form, that's a signal that they still have some use for the form. Otherwise, it will sit in the archived state for a while and eventually be deleted.


db/migrate/20240927200908_add_archived_at_to_form.rb
```diff
@@ -0,0 +1,23 @@
+class AddArchivedAtToForm < ActiveRecord::Migration[7.1]
+  def up
+    add_column :forms, :archived_at, :datetime
+
+    execute <<-SQL.squish
+      WITH most_recent_form_archives AS (
```
sanason, line 6: This update executed in ~3 seconds in my production-sized local version of the db.

MaroyaF (approved):

spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'rails_helper'
+
+RSpec.describe ApplicationHelper, type: :helper do
```
sanason, line 3: Moved tests to fiscal_year_spec.rb.

app/controllers/admin/record_retention_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    def export_form_managers
+      form_manager_emails = Form.forms_whose_retention_period_has_passed
+                        .flat_map(&:form_managers)
+                        .map(&:email)
+                        .sort
+                        .uniq
+      csv_content = CSV.generate(headers: false) do |csv|
+        form_manager_emails.each { |email| csv << [email] }
+      end
+      send_data csv_content, filename: "record-retention-form-manager-emails-#{timestamp_string}.txt"
```
ryanwoldatwork, line 24: maybe .csv extension?

ryanwoldatwork (approved):

app/controllers/admin/record_retention_controller.rb
sanason: Agreed, changed.

PR #1658, opened 2024-10-17
Support multiple Touchpoints surveys on a single page

sanason:
One of our customers has 2 Touchpoints surveys on [the same page](https://oig.hhs.gov/reports/recommendations/tracker). The new USWDS-based modal doesn't work in this case because each copy of the survey calls `Modal.on();` when it loads and each such call calls `init` on all Touchpoints modals loaded in the page, so `init` is called twice on whichever modal loads first. The USWDS code tends to break whenever `init` is called multiple times on the same component.

This PR solves the multiple-init problem by limiting the scope of each `Modal.on()` call to just the modal that is currently being loaded.

Manual tests performed:

- [x] **Two modal surveys on same page works:** I created a form with a combobox and date picker and set delivery method to "Blue button & modal". I copied the form and set the copy's delivery method to "Custom button & modal". I created a test HTML page containing both forms and tested that both modals work, both comboboxes work and both date pickers work.
- [x] **Modal survey works on a page that has its own copy of USWDS:** Taking the test page and surveys from the previous test, I added a script tag to the test page `<script src="https://cdnjs.cloudflare.com/ajax/libs/uswds/3.5.0/js/uswds.min.js" integrity="sha512-gXohKcsWNIdQYbKGl6kELTJlXd5RSZfNY1g+WRASW7Uu6SFr3fi2TDSk94JnBUP+wZPDWSYjzTJrQIc8/3NYHA==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>` and checked that both survey modals still work.
- [x] **Modal survey works on a page that has its own USWDS modal:** Taking the test page and surveys from the previous test, I added a USWDS modal to the page and tested that the page's modal plus both of the survey modals work.


gulpfile.js
```diff
@@ -42,6 +42,7 @@ embeddedWidgetPath = './app/views/components/widget';
 async function bundleWidgetJS() {
   return browserify("uswds/widget-uswds.js", {
     paths: ['./node_modules'],
+    standalone: 'fbaUswds'
```
sanason, line 45: The standalone option is required for Browserify to create a bundle such that the code inside the bundle can be accessed from code outside the bundle.

https://stackoverflow.com/questions/23296094/browserify-how-to-call-function-bundled-in-a-file-generated-through-browserify

ryanwoldatwork (approved):
👍

PR #1662, opened 2024-10-21
Build widget JS bundle for browser-only instead of using UMD

sanason:
Fixes a bug introduced in https://github.com/GSA/touchpoints/pull/1658.

In that PR, I changed the USWDS JS bundle that we use in the widget with the goal of making the modules `Modal`, `ComboBox` etc. accessible outside the bundle. I used the Browserify `standalone` option to do this. Standalone mode generates a UMD bundle that in theory can be used in the browser or with a variety of module systems. But the bundle I generated apparently didn't work with the module system on https://www.ncei.noaa.gov/maps/crn/.

This PR simplifies the situation by assuming that the generated bundle will only be used in the browser which means that the USWDS code can be exposed simply by hanging it off `window`. That is, in fact, how [USWDS suggests](https://designsystem.digital.gov/documentation/developers/#js-customization-2) that its components be exposed.

Still to do - needs manual testing.


ryanwoldatwork (approved):
i tested manually, adding a custom-button and inline form to the standard button preview page

👍

PR #1674, opened 2024-10-24
Deprecate v1 cx collections

ryanwoldatwork:
the PR removes the v1 CX Collections features.
v2 has been active for a year now.

https://github.com/GSA/touchpoints/pull/1674


config/initializers/us_timezones.rb
```diff
@@ -0,0 +1,7 @@
+US_TIMEZONES = [
```
ryanwoldatwork, line 1: moved timezones into an initializer to make them easier to re-use (compared to being in ApplicationHelper)

and, this returns an array of arrays, with the timezone itself, plus it's `name`. this is ultimately to fix a bug when the timezone string was used with `formatted_time`.  the `name` is what is expected.

there's a test below for the regression, which i'll comment on too

spec/features/admin/forms_spec.rb
```diff
@@ -751,6 +751,19 @@
           end
         end
 
+        describe 'editing form timezone' do
+          before do
+            select(US_TIMEZONES.sample[1], from: 'form_time_zone')
```
ryanwoldatwork, line 756: here's a test to ensure the timezone selected for a form (from the UI) is valid, by checking that the submissions (which use `formatted_time` still load properly).

app/views/user_mailer/user_welcome_email.html.erb
```diff
@@ -4,7 +4,7 @@
 </p>
 <p>
   We received a notification that you recently logged into Touchpoints
-  so I wanted to introduce myself and let you know our Feedback Analytics team is available
+  so we wanted to welcome and let you know our Feedback Analytics team is available
```
ryanwoldatwork, line 7: minor text updates to make i/we consistent

app/views/admin/organizations/show.html.erb
```diff
@@ -198,17 +200,17 @@
...
       <h3>
-        <%= pluralize(@collections.size, "Collection") %>
+        <%= pluralize(@cx_collections.size, "Collection") %>
       </h3>
 
-      <%- if @collections.present? %>
+      <%- if @cx_collections.present? %>
         <ul class="usa-list">
-        <% @collections.order(:year, :quarter).each do |collection| %>
+        <% @cx_collections.order(:year, :quarter).each do |cx_collection| %>
```
sanason, line 208:
```suggestion
        <% @cx_collections.order(:fiscal_year, :quarter).each do |cx_collection| %>
```
Screenshot 2024-11-05 at 8 00 42 PM

ryanwoldatwork: thanks!

also added a spec

sanason (approved):

PR #1676, opened 2024-10-25
show a table of services

ryanwoldatwork:
* rather than a list

### DIFF
--- a/app/views/admin/services/_results.html.erb
+++ b/app/views/admin/services/_results.html.erb
@@ -11,7 +11,7 @@
     </tr>
   </thead>
   <tbody>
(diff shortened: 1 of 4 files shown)


app/views/admin/services/_results.html.erb
```diff
@@ -11,7 +11,7 @@
     </tr>
   </thead>
   <tbody>
-    <% @services.each do |service| %>
+    <% services.each do |service| %>
```
ryanwoldatwork, line 14: `@services` relies on the instance variable being available from the controller,
whereas `services relies on the variable being passed into the partial

sanason (approved):

PR #1677, opened 2024-10-25
USDR account to service

ryanwoldatwork:
this PR deprecates the duplicative 'account' field which is/should be the same as 'service' field

note: the USDR may be deprecated as well, given M-17-06 was rescinded by M-23-22

### DIFF
--- a/app/models/digital_service_account.rb
+++ b/app/models/digital_service_account.rb
@@ -14,7 +14,7 @@ class DigitalServiceAccount < ApplicationRecord
 
   validates :name, presence: true
   validates :name, uniqueness: { scope: :account }
-  validates :account, presence: true
+  validates :service, presence: true
   validate :validate_account_types
   validates :service_url, presence: true
   validates :service_url, uniqueness: true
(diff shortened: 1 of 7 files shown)


public/api/v1/openapi.yml
```diff
@@ -836,8 +836,6 @@
         properties:
           name:
             type: "string"
-          account:
```
sanason, line 839: The `DigitalServiceAccount` type doesn't have a `service` field. Should that be added?

app/models/digital_service_account.rb
```diff
@@ -13,24 +13,24 @@ class DigitalServiceAccount < ApplicationRecord
   belongs_to :organization, optional: true
 
   validates :name, presence: true
-  validates :name, uniqueness: { scope: :account }
-  validates :account, presence: true
```
sanason, line 17: Any reason not to drop the `account` column in the db?

ryanwoldatwork: can do that as a follow-up

ryanwoldatwork: added

public/api/v1/openapi.yml
ryanwoldatwork: added

sanason (approved):

PR #1705, opened 2024-11-22
Index by id

ryanwoldatwork:


db/migrate/20241122185330_index_ids.rb
sanason, line 1: Doesn't Postgres automatically put an index on a table's primary key? 
https://www.postgresql.org/docs/17/sql-createtable.html#SQL-CREATETABLE-PARMS-PRIMARY-KEY

ryanwoldatwork: yes

PR #1752, opened 2025-01-17
bulk actions on submissions

ryanwoldatwork:
enables users to bulk select responses and a) flag them or b) mark them as archived.

later, additional options may be added, like: mark as spam, or assign tags, in bulk


sanason (approved):
Nice!

PR #1760, opened 2025-01-30
Rm early submission

ryanwoldatwork:
remove "early submission", a feature from the version 1 of the A-11 form, where a multi-page form can be optionally submitted at the end of the first page.

this was designed for the v1 A-11 form, that had 2 mandatory questions on page 1, and at least 5 additional (sometimes optional) questions on page 2.

the a-11 v2 (thumbs up/down) has been out for quite sometime, and this is a step moving forward with the deprecation of the v1 form.


sanason (dismissed):
Are there no currently active forms that are using early submission?

ryanwoldatwork:
yes, there are, but these would be addressed prior to deployment.

sanason:
Are there any using it that aren't A-11 forms? Maybe early submission is a more generally useful feature.

ryanwoldatwork:
Will keep an eye out.

I've went through a handful earlier today, and many forms have the early_submission box checked because it was copied from the template that started out with it, and then form template was edited as to not even have multiple sections/pages.

PR #1766, opened 2025-02-03
mark a submission as spam

ryanwoldatwork:
This PR allows for a submission to be marked as `spam` manually by a user.
This can be done a) in bulk or b) for an individual submission.

This PR also changes the submissions table, by not showing every response in the table (which often leads to ugly, horizontally scrollable tables, because they're wide and show many questions).  Rather, this view only shows columns that have "name" and/or "email" in the name.

The goal of this is to get staffers to process each feedback item on its own - like an email needs individualized attention.

Otherwise, the CSV export and API remain the recommended way to do deeper, custom analysis.

## Before

Screenshot 2025-02-06 at 9 16 26 AM

### After

Screenshot 2025-02-06 at 9 23 23 AM

Update:

image


sanason (approved):
👍

PR #1901, opened 2025-11-10
Some fixes to Cron

pburkholder:
Trying to address a few problems with this PR:

1. Running all the tasks in sequence without waiting for them resulting in a big spike in resource consumption, hitting quota, and the failing the last run-tasks
2. Using `--wait` apparently uncovers some failures that were masked when we just kicked off the task and hoped for the best
3. Using `set -e` would cause the later tasks not to run if there was a failure earlier.

### DIFF
--- a/.circleci/cron.sh
+++ b/.circleci/cron.sh
@@ -2,7 +2,7 @@
 
 # Fail if anything within this script returns
 # a non-zero exit code
-# set -e
+set -e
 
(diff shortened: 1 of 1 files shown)


sanason (approved):
It's been months since the Touchpoints cron job ran successfully. Let's give this a try. Thanks, @pburkholder !

PR #1958, opened 2026-01-30
Split deployment workflow into build-test-deploy jobs.

sanason:
Currently, the deployment workflow is a single CircleCi job with many steps, which means that, if any step fails, the entire workflow must be started from the beginning. So, for instance, if the tests pass but something transitory causes the deployment to cloud.gov to fail, you can't just restart the cloud.gov deployment piece, first you have to rerun all the tests again.

This PR splits the workflow into 3 sequential jobs - build, test, deploy. Now, if a job fails, the workflow only needs to be re-run from the start of the failed job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,37 +12,12 @@ version: 2.1
 jobs:
   build:
     docker:
-       - image: cimg/ruby:3.2.8-browsers # Matches deployed Ruby version in CF
-         environment:
-            RAILS_ENV: test
-            PGHOST: 127.0.0.1
-            PGUSER: root
-
-       - image:  cimg/redis:7.2.7
-         environment:
-            REDIS_URL: redis://redis:6379/1
-
-      # Specify service dependencies here if necessary
-      # CircleCI maintains a library of pre-built images
-      # documented at https://circleci.com/docs/2.0/circleci-images/
-       - image: cimg/postgres:15.8
-         environment:
-            POSTGRES_USER: root
-            POSTGRES_DB: touchpoints_test
-
-    parallelism: 4
-    working_directory: ~/repo
+       - image: cimg/ruby:3.2.8-node # Matches deployed Ruby version in CF
 
     steps:
       - run:
           name: Update packages
           command: sudo apt-get update
-
-      - run:
-          name: Ensure Chrome is available
-          command: |
-            # cimg/ruby:*-browsers images already include Chrome; skip orb command to avoid "Cannot find declaration" errors
-            echo "Using cimg/ruby:3.4.7-browsers which includes Chrome"
 
       - checkout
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -94,25 +69,59 @@ jobs:
...
+          key: v3-bundle-{{ checksum "Gemfile.lock" }}
 
       # Precompile assets (to use npm packages specified in assets.rb)
       - run:
           name: Precompile assets
           command: |
             npm i
-            rails assets:precompile
+            bundle exec rails assets:precompile
+
```
sanason: Copy the entire working directory to be used in later jobs.

.circleci/config.yml
```diff
@@ -94,25 +69,59 @@ jobs:
...
       - restore_cache:
           keys:
-          - v1-bundle-{{ checksum "Gemfile.lock" }}
+          - v3-bundle-{{ checksum "Gemfile.lock" }}
 
-      - run: bundle install
+      - run:
+          name: Install gems
+          command: |
+            bundle config set deployment true
```
sanason, line 77: 'set deployment true' ensures that the gems will be installed _within_ the working directory, so that they may be copied as part of the workspace.

.circleci/config.yml
```diff
@@ -142,11 +157,6 @@ jobs:
       - run:
           name: Deploy Sidekiq worker servers
           command: |
-            # Only deploy from a single parallel node to avoid concurrent CF pushes.
-            if [ "${CIRCLE_NODE_INDEX:-0}" != "0" ]; then
-              echo "Skipping Sidekiq deploy on parallel node ${CIRCLE_NODE_INDEX}"
-              exit 0
-            fi
```
sanason, line 149: Since the deploy job is not parallelized, I don't think we need this guard anymore.

.circleci/config.yml
```diff
@@ -12,38 +12,13 @@ version: 2.1
...
       - run:
           name: Update packages
           command: sudo apt-get update
 
-      - run:
-          name: Ensure Chrome is available
-          command: |
-            # cimg/ruby:*-browsers images already include Chrome; skip orb command to avoid "Cannot find declaration" errors
-            echo "Using cimg/ruby:3.4.7-browsers which includes Chrome"
-
```
sanason, line 46: I don't think we need warning this anymore.

.circleci/config.yml
```diff
@@ -94,25 +69,59 @@ jobs:
...
           name: Precompile assets
           command: |
             npm i
-            rails assets:precompile
+            bundle exec rails assets:precompile
+
+      - persist_to_workspace:
+          root: .
+          paths:
+            - .
```
sanason, line 95: Copy the entire working directory to be used in later jobs.

PR #1969, opened 2026-02-05
Fix duplicate CORS header issue.

sanason:
This PR fixes a CORS error that a couple of our users reported seeing for their embedded Touchpoints form. The error is currently visible on https://touchpoints.digital.gov too:

Screenshot 2026-02-05 at 2 04 24 PM

Looking at the network request, the problem appears to be that CORS headers are being added twice to the response:
Screenshot 2026-02-05 at 2 08 33 PM

Looking at the code, I identified the 2 places where CORS headers are added. The duplication only happens for static files in `public/` that are served by the `ActionDispatch::Static`:

1. Headers are added by the `cors-rack` gem in [application.rb](https://github.com/GSA/touchpoints/blob/fd207a316e4194a12dd7f7bb02348bdd8650f8ef/config/application.rb#L31).
2. Headers are added by configuration in [production.rb](https://github.com/GSA/touchpoints/blob/fd207a316e4194a12dd7f7bb02348bdd8650f8ef/config/environments/production.rb#L26).

Since the `cors-rack` code is more general-purpose and has been around longer, I kept it and removed the configuration code.


config/environments/production.rb
```diff
@@ -18,15 +18,12 @@
   # Do not fall back to assets pipeline if a precompiled asset is missed.
   config.assets.compile = false
 
-  # Let Cloud Foundry / container platforms serve precompiled assets from /public.
-  config.public_file_server.enabled = ENV["RAILS_SERVE_STATIC_FILES"].present?
```
sanason, line 22: I don't see that it's useful to have this configurable at deploy time (switching to use NGINX/Apache instead of the Rails server would require additional code changes anyway), and hard-coding it means you don't have to look in 2 places to understand the configuration.

PR #1984, opened 2026-02-27
Release to production

sanason:
Changes:

- Several Dependabot updates
- Disabled Bullet alerts in development mode (only affects Touchpoints developers)
- Admins no longer see all forms in My Forms (only affects Touchpoints admins)
- Upgrade to Ruby 3.2.10


sanason:
I have manually smoke-tested these changes on staging.

PR #1995, opened 2026-03-09
Don't validate submissions during deletion.

### LINKED ISSUE #1990: Bug: Survey response deletion appears to succeed but response is not deleted

sanason:
When a survey response is submitted, Touchpoints runs model validations checking the submitted answers against the corresponding survey questions - an email question must have an email answer, a required question must have a non-nil answer, a dropdown question must have an answer that is one of the dropdown options, etc. This is sensible and useful for rejecting spam responses, among other uses.

Does this validation need to run again at any point after the response has been submitted? A submitted response is largely immutable but it does have a set of status flags can be changed by the survey owners as they process responses. For instance, responses can be flagged, tagged, archived, deleted, etc. Currently, Touchpoints runs the question-answer validations during these status updates too. This causes a problem because, although the answers cannot change after the response is created, the form and its questions can change, so a response that had valid answers when it was submitted might not be valid at a later date. In the case of delete and #1990 , Touchpoints was silently swallowing such validation errors, giving the impression that it was deleting responses when it was not.

I say there is no reason to re-run question/answer validation when it is only the status of a response that has changed. This PR updates the response model to only run such validations on submission.

### DIFF
--- a/app/controllers/admin/submissions_controller.rb
+++ b/app/controllers/admin/submissions_controller.rb
@@ -186,23 +186,20 @@ module Admin
     def delete
       ensure_form_manager(form: @form)
 
-      if @submission.update(deleted: true, deleted_at: Time.now)
-        Event.log_event(Event.names[:response_deleted], 'Submission', @submission.id, "Submission #{@submission.id} deleted at #{DateTime.now}", current_user.id)
-      else
-        Rails.logger.warn("Failed to delete submission: #{@submission.errors.full_messages.join(', ')}")
-      end
+      Event.log_event(Event.names[:response_deleted], 'Submission', @submission.id, "Submission #{@submission.id} deleted at #{DateTime.now}", current_user.id)
+      @submission.assign_attributes(deleted: true, deleted_at: Time.now)
+      @submission.save(validate: false)
     end
 
     def destroy
       ensure_form_manager(form: @form)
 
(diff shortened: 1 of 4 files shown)


app/controllers/admin/submissions_controller.rb
```diff
@@ -186,31 +186,29 @@ def unmark
     def delete
       ensure_form_manager(form: @form)
 
-      if @submission.update(deleted: true, deleted_at: Time.now)
-        Event.log_event(Event.names[:response_deleted], 'Submission', @submission.id, "Submission #{@submission.id} deleted at #{DateTime.now}", current_user.id)
-      else
-        Rails.logger.warn("Failed to delete submission: #{@submission.errors.full_messages.join(', ')}")
-      end
+      Event.log_event(Event.names[:response_deleted], 'Submission', @submission.id, "Submission #{@submission.id} deleted at #{DateTime.now}", current_user.id)
+      @submission.assign_attributes(deleted: true, deleted_at: Time.now)
```
sanason, line 190: Do we really need to disable validation here, given that we set `validate_answers` to only run during submission creation?

No, but in case anyone adds additional validation routines in the future, I want avoid a repeat of the silent swallowing of validation failure plus skipping validation is consistent with the other status flags updates in this controller that all use the non-validating `update_attribute` call.

PR #2014, opened 2026-04-01
Delete unused Github workflows

sanason:
I was going to update this file to pin the GitHub action SHA's but then started to wonder whether the file could just be deleted, since we use CircleCI. @rileyseaburg, opinion?


rileyseaburg:
Yes go ahead!

PR #2066, opened 2026-05-23
Add form responses api endpoint

### LINKED ISSUE #2042: Touchpoints API limit
We have multiple sites at NOAA that exceed the 5000 API cap. We were wondering if it was possible on a case by cases basis to expand the cap for specific accounts.

sanason:
Adding a new API endpoint `GET /api/v1/forms/{:id}/responses`. Currently, the way to get responses from a form is to call `GET /api/v1/forms/{:id}` and extract the responses from the returned form object. There are a few bugs with this method, however:

1. Deleted responses are included.
2. Pagination does not work correctly when the responses are filtered by date.
3. Pagination links returned by the endpoint refer to the raw API endpoints on the Touchpoints server (which are inaccessible to the public) instead of to the api.data.gov gateway.

The new endpoint introduced in this PR fixes the first and second bugs. The 3rd bug will be fixed in a subsequent PR.

This PR does not try to fix the existing endpoint. I plan to deprecate the `submissions` field that is returned by `GET /api/v1/forms/{:id}`.

### DIFF
--- a/app/controllers/api/v1/forms_controller.rb
+++ b/app/controllers/api/v1/forms_controller.rb
@@ -54,6 +54,15 @@ module Api
       end
 
       def responses
+        # Start temporary code for debugging in staging
+        Rails.logger.info "--- REQUEST HEADERS ---"
+        request.headers.each do |key, value|
+          # Filter for standard HTTP headers if preferred
+          Rails.logger.info "#{key}: #{value}"
+        end
+        Rails.logger.info "-----------------------"
+        # End temporary code
+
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/forms_controller.rb
```diff
@@ -53,6 +53,64 @@ def show
         end
       end
 
+      def responses
```
sanason, line 56: This temporary code is here so I can figure out which headers we receive on requests forwarded from api.data.gov. It should not be deployed to production.

PR #2072, opened 2026-05-28
Fix bugs in email domain matching to map users to organizations

### LINKED ISSUE #2035: Validate that organization domain is bare domain name
When creating or editing an organization in Touchpoints, the 'Domain' field should validate that a domain name has been entered and not, for instance, a URL. This field controls which user accounts are considered to belong to a given organization (which controls a lot of what the user can see), so it's important that it's filled in correctly.

**Correct**

Image

**Incorrect (but permitted by form)**

Image

### LINKED ISSUE #2058: [Bug] Users with four-level email domains are not assigned to correct organization
Example:

The FEMA organization is configured such that Login.gov users with the 'fema.dhs.gov' email suffix are supposed to be assigned to FEMA. This works for 'john.smith@fema.dhs.gov'. But it doesn't work for 'jill.smith@associates.fema.dhs.gov'. Jill gets assigned to the DHS organization instead of to FEMA, because if the full domain does not match exactly, the code falls back to the second-level domain to identify the user's organization.

sanason:
Fixes the 2 linked bugs.

For organizations, the PR adds validation for the `domain` field:

- Domains must be unique. A user is assigned to a unique organization by finding the organization whose `domain` most closely matches the domain of the user's email address. If multiple organizations have the same domain, there's no way to guarantee which organization a user will be assigned to.
- Domains must be entered as actual domain strings, as enforced by a regex. Previously, the domain field was a free-form text field which sometimes was filled with URL's or other inappropriate string.
- Domains are limited to 5 parts (e.g., 'sub1.sub2.sub3.example.gov'), in order to put an upper bound on the number of look-ups we have to do when deciding which organization a new user belongs to. This limit covers all current organizations.

For users, the PR fixes the logic used to assign a new user to an existing organization. Conceptually, the logic is now:

1. Take the user's email address as passed in by Login.gov and extract the email's domain.
2. Find all organizations whose domain is a suffix of the user's email domain (e.g., organizations 'dhs.gov' and 'fema.dhs.gov' both match jill.smith@associates.fema.dhs.gov.)
3. Choose the organization with the longest domain (e.g. 'fema.dhs.gov' in this case).

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -55,46 +55,6 @@ RSpec.describe User, type: :model do
       end
     end
 
-    context 'with a valid .gov email address (under 1 subdomain) for an Organization that has been created' do
-      before do
-        FactoryBot.create(:organization)
-        @user.email = 'user@associates.subdomain.example.gov'
-        @user.save
-      end
-
-      it 'is valid' do
-        expect(@user.valid?).to eq(true)
-        expect(@user.persisted?).to eq(true)
-      end
-    end
-
-    context 'with a valid .gov email address (under 2 subdomains) for an Organization that has been created' do
-      before do
-        FactoryBot.create(:organization)
-        @user.email = 'user@associates.subdomain.example.gov'
-        @user.save
-      end
-
-      it 'is valid' do
-        expect(@user.valid?).to eq(true)
-        expect(@user.persisted?).to eq(true)
-      end
-    end
-
-    context 'with a valid .gov email address (under 3 subdomains) for an Organization that has been created' do
-      before do
-        FactoryBot.create(:organization)
-        @user.email = 'user@toomany.associates.subdomain.example.gov'
-        @user.save
-      end
-
-      it 'fails tld_check' do
-        expect(@user.valid?).to eq(false)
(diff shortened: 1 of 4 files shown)


spec/models/user_spec.rb
```diff
@@ -55,46 +55,6 @@
...
-        @user.save
-      end
-
-      it 'is valid' do
-        expect(@user.valid?).to eq(true)
-        expect(@user.persisted?).to eq(true)
-      end
-    end
-
-    context 'with a valid .gov email address (under 3 subdomains) for an Organization that has been created' do
```
sanason, line 84: This test would fail now. Now, there's no limit on how many segments are in the user's email domain, but we do only match against the last 5 since organization domains are now limited to 5 segments.

spec/models/user_spec.rb
```diff
@@ -55,46 +55,6 @@
       end
     end
 
-    context 'with a valid .gov email address (under 1 subdomain) for an Organization that has been created' do
```
sanason, line 58: This test and the next one are redundant, given the tests under '#ensure_organization'.

PR #2074, opened 2026-06-01
Add a way to remove a logo from a form

### LINKED ISSUE #1972: [Bug] Provide a way to remove a logo from a form
The UI for adding a logo to a form provides no way to later remove the logo. Fix this.

Image

sanason:
Currently, the UI for managing a form's logo looks like this (this screenshot is from a form that already has a logo added):

Screenshot 2026-06-03 at 11 48 34 AM

It's possible to change the logo file (by uploading a new file in the file input) and it's possible to change the display format of the logo (selecting between banner and square) but there is no way to go back to not having any logo.

This PR updates the UI to add a 'Delete Logo' button:
Screenshot 2026-06-03 at 11 55 25 AM

That's the main change. A few other smaller changes:

- Moved the image of the current logo to the top so that buttons would be at the bottom of the form.
- Added a file type hint to the 'Upload new logo' label
- Added error handling so that a failed upload displays an error instead of failing silently.


config/locales/en.yml
```diff
@@ -4,6 +4,19 @@ en:
     request:
       unauthorized_host: "request made from non-authorized host"
     messages:
```
sanason, line 6: Copied from the Carrierwave documentation [here](https://github.com/carrierwaveuploader/carrierwave#i18n).

public/api/v1/openapi.yml
```diff
@@ -251,11 +251,10 @@
                 - "whitelist_url_8"
                 - "whitelist_url_9"
                 - "whitelist_test_url"
```
sanason, line 253: I should not be making breaking changes to the API but I think it unlikely that people are using these fields in the API response. Perhaps I will be proved wrong.

app/views/components/forms/edit/_builder.html.erb
```diff
@@ -2,14 +2,16 @@
 <% @disclaimer_text_limit = 1000 %>
 <%= javascript_include_tag "quill.js" %>
 <%= stylesheet_link_tag "quill-snow.css" %>
-<%= render 'admin/forms/logo_display', { form: form } %>
+<div id="display-logo-div">
```
sanason, line 5: Moving the containing `div` out of the partial so that `update_display_logo.js.erb` is not adding a new nested `div` every time it runs.

PR #2085, opened 2026-06-24
Use rswag to update openapi spec

### LINKED ISSUE #2061: Improve Touchpoints API documentation

sanason:
The purpose of this PR is to update the OpenAPI spec for the V1 Touchpoints API. The updated spec adds documentation for the following endpoints:

- /forms/{id}/responses
- /service_providers
- /services
- /services/{id}
- /cx_collection_details
- /cx_collections
- /cx_responses

It removes documentation for two digital registry endpoints:

- /digital_products
- /digital_service_accounts

Those endpoints still exist but, since the public-facing digital registry was shuttered in 2024, I'm not willing to maintain the documentation for the corresponding API endpoints. /websites is still documented because I know that GSA is still using the website registry as its internal inventory tool.

### Adding a tool to automate generation of API documentation
I'm not sure how the OpenAPI spec was maintained previously but there weren't any spec automation libraries listed in the Gemfile. I adopted the [Rswag](https://github.com/rswag/rswag) tool to help with automation. Setting up Rswag is how this PR got so big.

With the addition of Rswag, the V1 spec file (public/api/v1/openapi.yml) becomes a generated file. It is generated by the command:
```
RSWAG_DRY_RUN=0 bundle exec rake rswag:specs:swaggerize
```
It's still a version controlled file because changes to it should be reviewed since it is human-facing documentation. ~~Since it is a generated, version-controlled file, I added a CI step to check that developers are re-running the generation script before commits so the spec does not drift from the actual API.~~

Rswag doesn't really reduce the manual effort required to create the OpenAPI spec. But it helps the developer to write _accurate_ documentation because it runs tests that actually call the API and validate the request/response against what the documentation says.

### Other miscellaneous changes

- Cleaned up a couple of abandoned API endpoints
- Set the entire V1 API to be `json` by default. It's no longer required to add `.json` to the end of the URL in your API calls (though it still works if you do).

_AI-assisted: authored with an AI coding agent under human review._


app/controllers/api/v1/omb_cx_reporting_collections_controller.rb
```diff
@@ -1,19 +0,0 @@
-# frozen_string_literal: true
-
```
sanason, line 2: Safe to delete - no routes are mapped to this controller.

app/controllers/api/v1/cx_responses_controller.rb
```diff
@@ -27,6 +27,7 @@ def index
               meta: {
                 current_page: cx_responses.current_page,
                 size: cx_responses.size,
```
sanason, line 29: Adding this here so it matches the `meta` block in `/forms/{id}/responses`.

app/controllers/api/v1/forms_controller.rb
```diff
@@ -8,12 +8,10 @@ def index
           format.json do
             if current_user.organizational_admin?
               render json: current_user.organization.forms
-                .order(:id)
-                .limit(100), each_serializer: FormSerializer
```
sanason, line 12: I don't like this undocumented and arbitrary limit. A safer solution would be to add pagination but the API already has a bunch of unlimited endpoints and I don't think this one is more of a risk than any of them.

config/routes.rb
```diff
@@ -37,13 +38,12 @@
...
   end
 
-  namespace :api do
+  namespace :api, defaults: { format: :json } do
     namespace :v0 do
       resources :forms, only: %i[index show]
     end
     namespace :v1 do
       resources :organizations, only: [:index]
-      resources :collections, only: [:index]
```
sanason, line 46: Deleting because this route has no corresponding controller.

spec/models/cx_collection_detail_spec.rb
```diff
@@ -54,8 +54,8 @@
         expect(csv.class).to eq(CSV::Table)
         expect(csv.size).to eq(2)
```
sanason, line 56: Changed because the factory changed.

spec/factories/cx_collection_detail_upload.rb
```diff
@@ -9,27 +9,27 @@
...
-          negative_humanity: [true, false].sample,
-          negative_employee: [true, false].sample,
-          negative_other: [true, false].sample,
+          question_1: ['1', '0'].sample,
+          positive_effectiveness: ['1', '0'].sample,
+          positive_ease: ['1', '0'].sample,
+          positive_efficiency: ['1', '0'].sample,
+          positive_transparency: ['1', '0'].sample,
+          positive_humanity: ['1', '0'].sample,
+          positive_employee: ['1', '0'].sample,
```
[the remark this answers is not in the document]
sanason: Not a problem. This is just sample data generated for testing.

PR #2091, opened 2026-07-10
Minor changes to openapi.yaml spec

sanason:


public/api/v1/openapi.yml
```diff
@@ -27,10 +22,8 @@ tags:
 - name: CX Collections
 - name: Digital Registry
 servers:
```
sanason, line 24: Note: This file was produced by running the rswag generator and then manually deleting the localhost entry from `servers`. I don't want localhost listed as the first/default server in our official API documentation but I haven't found a way yet to get the RSwag tests to pass without listing localhost first in `swagger_helper.rb`. Obviously, it's bad to edit auto-generated files. I will automate this step someday.

PR #2102, opened 2026-07-23
Fix scroll to top during page navigation in embedded forms

### LINKED ISSUE #2104: [Bug] Incorrect scroll behavior for multi-page embedded forms
Two issues:

- Multi-page forms deployed as modals do not scroll to the top when navigating between form pages.
- Multi-page forms deployed as inline scripts scroll to the top of the overall page, not to the top of the form, when navigating between form pages.

sanason:
See linked issue for problem description.

For hosted forms, we maintain the existing behavior, which is to scroll the window to the top of the page while navigating between pages of the form.

For modal forms, the new behavior is to scroll to the top of the modal while navigating between pages of the form. The scroll behavior is smooth, because that is the default scroll behavior for the USWDS modal.

For inline forms, the new behavior is to scroll to the top of the inline form while navigating between pages of the form.

### DIFF
--- a/app/views/components/widget/_fba.js.erb
+++ b/app/views/components/widget/_fba.js.erb
@@ -635,12 +635,11 @@ function FBAform(d, N) {
 					});
 					d.dispatchEvent(previousPageEvent);
 
-					// if in a modal, scroll to the top of the modal on previous button click
-					if(this.formComponent().getElementsByClassName("fba-modal")[0]) {
-						this.formComponent().scrollTo(0,0);
-					} else {
-						N.scrollTo(0, 0);
-					}
+          <%- if form.delivery_method == "touchpoints-hosted-only" %>
+          N.scrollTo(0, 0);
+          <% else %>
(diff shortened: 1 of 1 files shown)


app/views/components/widget/_fba.js.erb
```diff
@@ -635,12 +635,11 @@ function FBAform(d, N) {
 					});
 					d.dispatchEvent(previousPageEvent);
 
-					// if in a modal, scroll to the top of the modal on previous button click
-					if(this.formComponent().getElementsByClassName("fba-modal")[0]) {
-						this.formComponent().scrollTo(0,0);
```
sanason, line 640: This may have changed at some point but, currently, the scroll container is the parent of `this.formComponent()`, not the form component itself. `scrollTo` only works on a scroll container.

Switching to `scrollIntoView` makes the code more robust because `scrollIntoView` searches upward through an element's ancestors to find a scroll container.

PR #2105, opened 2026-07-24
Changes to Cloudflare Turnstile client-side implementation

sanason:
This PR makes a few client-side improvements to our implementation of Cloudflare Turnstile. We are testing out Turnstile on a few forms currently, both our own internal forms and some customer forms.

Changes:

1. For embedded forms, don't issue a challenge until the user has interacted with the form. 
**Justification:** Most users who visit the page containing an embedded form will never actually interact with the form. There's no reason to incur the overhead of issuing challenges unless the user indicates that they plan to submit the form. Even if they do eventually fill out the form, tokens have a limited lifespan and there's no reason to refresh the token over and over while waiting for the user to open the form.
Also, our Turnstile analytics are currently overwhelmed by an enormous number of challenges solved without any resulting form submission. I'd like to see numbers that only reflect humans (or bots) who have interacted with the form.
2. When the user is required to complete an interactive challenge, display a validation error if they try to submit the form without doing the challenge.
3. Auto-refresh the token when it expires.
4. Auto-refresh the interactive challenge when it expires.

### DIFF
--- a/app/views/components/forms/_custom.html.erb
+++ b/app/views/components/forms/_custom.html.erb
@@ -111,10 +111,7 @@
         <% end %>
         <%- if section == form.form_sections.last && !form.suppress_submit_button %>
           <%- if form.enable_turnstile? %>
-          <div class="margin-top-2">
-            <div id="turnstile-container"></div>
-            <%= hidden_field_tag "cf-turnstile-response", nil %>
-          </div>
+          <div id="fba-turnstile-container" class="fba-turnstile-container"></div>
           <% end %>
           <button type="submit" class="usa-button submit_form_button"><%= t 'form.submit' %></button>
         <% end %>
--- a/app/views/components/widget/_fba.js.erb
+++ b/app/views/components/widget/_fba.js.erb
@@ -210,22 +210,22 @@ function FBAform(d, N) {
 			this.resetErrors();
 			var formElement = this.formElement();
 			var self = this;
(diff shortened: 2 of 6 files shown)


app/views/components/forms/_custom.html.erb
```diff
@@ -111,10 +111,7 @@
         <% end %>
         <%- if section == form.form_sections.last && !form.suppress_submit_button %>
           <%- if form.enable_turnstile? %>
-          <div class="margin-top-2">
-            <div id="turnstile-container"></div>
-            <%= hidden_field_tag "cf-turnstile-response", nil %>
```
sanason, line 116: Turnstile manages its own hidden form field so this `hidden_field_tag` isn't necessary.

app/views/components/widget/_fba.js.erb
```diff
@@ -719,26 +725,80 @@ function FBAform(d, N) {
...
+    },
 		loadTurnstile: function() {
+      if (window.turnstile) {
+        <%# API already loaded (host page or another form loaded it) — render now. %>
+        this.initTurnstile();
+        return;
+      }
 			let script = document.createElement("script");
-			script.src = "https://challenges.cloudflare.com/turnstile/v0/api.js";
-			script.async = true;
```
sanason, line 725: Dynamically inserted scripts are automatically asynchronous.

PR #2114, opened 2026-08-04
Small fixes to turnstile widget

sanason:
Two small fixes to the Turnstile widget.

### DIFF
--- a/app/views/components/widget/_fba.js.erb
+++ b/app/views/components/widget/_fba.js.erb
@@ -589,7 +589,7 @@ function FBAform(d, N) {
 			xhr.send(JSON.stringify({
 				<%- if form.enable_turnstile? %>
         <%# This hidden form field is automatically created and managed by Turnstile. %>
-				"cf-turnstile-response" : form.querySelector("input[name='cf-turnstile-response']") ? form.querySelector("input[name='cf-turnstile-response']").value : null,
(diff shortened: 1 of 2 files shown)


app/views/components/widget/_fba.js.erb
```diff
@@ -589,7 +589,7 @@ function FBAform(d, N) {
 			xhr.send(JSON.stringify({
 				<%- if form.enable_turnstile? %>
         <%# This hidden form field is automatically created and managed by Turnstile. %>
-				"cf-turnstile-response" : form.querySelector("input[name='cf-turnstile-response']") ? form.querySelector("input[name='cf-turnstile-response']").value : null,
```
sanason, line 592: Snake case variable for consistency.

app/views/components/widget/_fba.js.erb
```diff
@@ -796,7 +797,7 @@ function FBAform(d, N) {
     },
     validateTurnstile: function() {
       this.clearTurnstileValidationError();
-      const turnstileToken = turnstile.getResponse(self.turnstileWidgetId);
```
sanason, line 799: `self` is not defined here. This line still worked because Turnstile searches for a widget if you pass `undefined`, but you're supposed to call it with a widget id.

PR #2081, opened 2026-06-19
make alt text for form logo configurable

afomi:
https://github.com/GSA/touchpoints/issues/2078

PR #2116, opened 2026-08-05
Add spam_determination provenance to submissions

sanason:
## Context

Submissions can be marked as spam, but the `spam` boolean records only *that* a submission is spam, not *why* or *by whom/what*. Reviewers triaging the "marked as spam" bucket have no explanation to act on. This adds a structured provenance field, and is a precursor to introducing automated spam detection (a follow-up PR) that will populate the same field.

## What changed

- **Migration** — adds a nullable `jsonb` `spam_determination` column to `submissions` (no default, no index). Backfills existing spam rows with `{ "source": "manual" }` (all current spam rows were manually marked). Reversible (`up`/`down`).
- **Manual admin actions** — `mark`, `unmark`, and bulk `spam` set `spam_determination` to `{ "source": "manual" }` when marking and clear it (`nil`) when unmarking. Uses `update(..., validate: false)`, matching the prior `update_attribute` behavior (skips validation, runs callbacks, touches `updated_at`).
- **Model** — excludes `spam_determination` from the `validate_answers` submission-field check.
- **API** — `SubmissionSerializer` now exposes both `spam` and `spam_determination`; swagger schema (`swagger_helper.rb`) and the generated `public/api/v1/openapi.yml` updated to match. Note: `spam` was not previously serialized, so this is a new (additive) API field.
- **Admin UI** — the submission show page displays `spam_determination` in the metadata table.

### Provenance shape
- Manual: `{ "source": "manual" }`
- Automated (future PR): `{ "source": "automated", "checks": { "<check>": "<status>" } }`

Contains check names/statuses and source only — never free-text detail (avoids PII/cardinality risk).

## Verification

- `bundle exec rspec spec/models/submission_spec.rb spec/controllers/admin/submissions_controller_spec.rb` — passing (reported by author).
- Added specs: model (default nil, jsonb round-trip, not-an-invalid-field), controller (`mark`/`unmark`/bulk set/clear provenance).

## Rollback

Revert the commit and run `bin/rails db:rollback` to drop the column (migration `down` removes it).

## Security impact

Does not affect authentication or authorization. Adds a data field and begins exposing `spam`/`spam_determination` via the API; contents are limited to check names/statuses and source (no PII). Admin write paths are unchanged in their authorization (`ensure_form_manager`).

## Notes for reviewers

- The `logo_alt_text:` line in `public/api/v1/openapi.yml` comes from a prior migration on this branch's base and is incidental to the regenerated spec.
- An ADR documenting this decision was drafted but intentionally left out of this PR.

---
🤖 This PR was prepared with AI assistance (OpenCode). All changes require human review before merge.

### DIFF
--- a/app/views/admin/submissions/show.html.erb
+++ b/app/views/admin/submissions/show.html.erb
@@ -186,14 +186,6 @@
           </tr>
           <tr>
             <td>
-              Spam prevention mechanism
-            </td>
-            <td>
-              <%= h(@submission.spam_prevention_mechanism) %>
-            </td>
-          </tr>
-          <tr>
-            <td>
               Language
             </td>
             <td>
@@ -214,6 +206,16 @@
             </td>
             <td>
               <%= @submission.spam? %>
+            </td>
+          </tr>
+          <tr>
+            <td>
+              Spam determination
+            </td>
+            <td>
+              <%- if @submission.spam_determination.present? %>
+                <%= h(@submission.spam_determination.to_json) %>
+              <% end %>
             </td>
           </tr>
           <tr>
--- a/public/api/v1/openapi.yml
+++ b/public/api/v1/openapi.yml
@@ -1087,6 +1087,17 @@ components:
         flagged:
           type: boolean
           description: Whether this submission has been flagged for review.
+        spam:
+          type: boolean
+          description: Whether this submission has been marked as spam (manually or
+            by automated spam detection).
+        spam_determination:
+          type: object
+          nullable: true
+          description: 'Provenance of the spam determination. For manual marks: {
+            source: "manual" }. For automated marks: { source: "automated", checks:
+            { <check_name>: <status> } }. Null when the submission is not marked as
+            spam.'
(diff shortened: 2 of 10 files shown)


public/api/v1/openapi.yml
```diff
@@ -1978,6 +1989,7 @@ paths:
                       whitelist_url_9:
                       whitelist_test_url: ''
                       header_logo_display: banner
```
sanason, line 1991: Forgot to run the API doc generator after https://github.com/GSA/touchpoints/pull/2081. `logo_alt_text` is unrelated to the current PR.

app/views/admin/submissions/show.html.erb
```diff
@@ -184,14 +184,6 @@
               <%= h(@submission.referer) %>
             </td>
           </tr>
-          <tr>
```
sanason, line 187: In the interest of not confusing the user with too many spam fields, I'm removing this from the view for the time being.

PR #2118, opened 2026-08-07
 Refactor spam checks into service objects and record metrics

sanason:
This PR introduces a centralized framework for detecting and handling spam form submissions. Having such a framework allows us to start collecting standardized metrics that we can use to compare the effectiveness of various spam detection methods. The framework is designed to support pluggable spam detection methods - if we have an idea for a new method, we can wire it in quickly and immediately start collecting metrics to prove whether or not it is effective.

### Metrics
Each form submission is run through all of our spam checks, which each return a verdict (one of pass, reject, flag, surface, not_applied). The individual verdicts are combined into an overall verdict for the submission. The result from each evaluation is sent to New Relic, where we will be able to compute the following metrics:

- How many total submissions passed, were rejected, were flagged, etc.
- How many times each spam check passed, rejected, flagged, etc.
- Correlations between spam checks - do spam check 1 and spam check 2 tend to return the same verdict for submissions

### Refactoring
Originally, spam checks were inserted in a somewhat ad hoc fashion throughout the controller code, which made it difficult to generate standardized metrics. So, this PR moves (some) of the existing spam checks into a centralized spam detection module. Since such a module doesn't fit well in either the controller or model layer, this PR also introduces a new service layer, which is justified in the project's first ADR.

### How spam checks have changed
No new spam checks are introduced by this PR. 3 existing checks have been fit into the framework - honeypot, referer validation and Cloudflare Turnstile. While the logic of the checks is the same, I did change how Touchpoints handles submissions that it suspects to be spam. Specifically, for each check:

- **Honeypot:** Previously, when the honeypot check failed, the submission call returned a generic `head 200` response, which I believe was supposed to conceal the fact that the submission had been blocked. But it was an imperfect concealment, since you get a different response if the submission actually succeeds. So, now we return the standard success response for submissions that fail honeypot (but the submission is not actually saved).
- **Referer validation:** Previously, when the referer validation check failed, the submission call returned an error. This check can fail for 2 reasons - it could be a bot or it could be because the form's allowlist is misconfigured. In either case, returning an error to the user is not useful. Now, on a failed referer validation check, we set the `spam` flag and save the submission and report success to the caller.
- **Turnstile:** Previously, for all turnstile errors, the submission call returned an error. Now, we distinguish between errors that the user can fix (expired token), which still return an error to the user, and errors that the user can't fix (Turnstile is down), for which we let the submission succeed.

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'uri'
-
 class SubmissionsController < ApplicationController
   before_action :set_form, only: %i[new create]
   append_before_action :verify_authenticity_token, if: :form_requires_verification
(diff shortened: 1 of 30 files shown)


app/services/spam_checks/referer.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
```
sanason, line 1: I have gone back to the original logic for the referer check, which I think is clearer than the logic introduced in https://github.com/GSA/touchpoints/pull/1898.

app/controllers/submissions_controller.rb
```diff
@@ -165,67 +165,19 @@ def set_form
   def submission_params
     permitted_fields = @form.questions.collect(&:answer_field)
     permitted_fields << %i[language location_code referer hostname page query_string fba_directive]
```
sanason, line 168: `cf_turnstile_response` is not part of the `submission` object - it comes as its own param - so this line doesn't do anything.

PR #2123, opened 2026-08-24
Embedded forms should not have h1 headings

### LINKED ISSUE #1954: Touchpoints modal adds an <h1> to the page, which likely leads to multiple <h1>s being on the page
The Touchpoints modal injects an `<h1>` into the DOM. For accessibility and SEO reasons, there should only be one `h1 `on the page, and the page likely already has an `<h1>` on it. See https://architecture.lullabot.com/adr/20250124-use-one-h1-per-page for some good reasons why.

**What should happen?**

Use an `<h2>` at least?

**How to reproduce:**

Click Feedback button, inspect the DOM of the modal feedback form contents, observe the `<h1>` there.

sanason:


app/views/admin/forms/example.html.erb
```diff
@@ -9,16 +9,98 @@
```
sanason, line 9: The example page got spiffed up a bit but has no fundamental changes. This is the page where you can preview how your embedded form will look when it is embedded on a webpage.

app/views/components/forms/_logo_and_title.html.erb
```diff
@@ -1,26 +1,27 @@
-<header>
```
sanason, line 1: In addition to making heading level configurable, I also moved the logo `div` outside of the heading tag, since div inside heading is not valid html.

PR #2125, opened 2026-09-01
Use render-context-specific heading tags in forms

### LINKED ISSUE #1954: Touchpoints modal adds an <h1> to the page, which likely leads to multiple <h1>s being on the page

sanason:
Fixes https://github.com/GSA/touchpoints/issues/1954.

Touchpoints forms use heading tags for two purposes:

1. The title of the form is a heading.
2. Various instances of the USWDS Alert component are used in forms and they may contain heading tags.

This PR adjusts the levels of these two types of heading tag based on how the form is being displayed:

- **Hosted on Touchpoints:** The title is `<h1>` and alert headings are `<h2>`.
- **Embedded as a modal:** The title is `<h2>` and alert headings are `<h3>`. (Note that there is [some disagreement](https://github.com/w3c/wcag/discussions/2722) about whether it is wrong to use `<h1>` for the title of a modal but `<h2>` seems to be the less controversial choice.)
- **Embedded as an inline form:** The title is `<h2>` and alert headings are `<h3>`. This fixes the bug in question (enforce only one `<h1>` on a page) but does nothing to guarantee that the form fits into the heading hierarchy of the overall page. I'm going to wait until someone complains before I try to fix the trickier 'overall hierarchy' problem.

### DIFF
--- a/app/controllers/touchpoints_controller.rb
+++ b/app/controllers/touchpoints_controller.rb
@@ -13,7 +13,7 @@ class TouchpointsController < ApplicationController
         js
       end
       format.css do
-        render(partial: 'components/widget/widget', formats: :css, locals: { form: @form })
+        render(partial: 'components/widget/widget', formats: :css)
       end
     end
   end
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -374,12 +374,6 @@ class Form < ApplicationRecord
 
   # Renders the widget CSS partial for use with the Rust widget renderer
   def render_widget_css
-    controller_with_request = build_controller_with_mock_request
-    controller_with_request.render_to_string(partial: 'components/widget/widget', formats: :css, locals: { form: self })
-  end
-
-  # Renders the widget CSS partial for use with the Rust widget renderer
-  def render_widget_css
     controller = ApplicationController.new
 
     # Set up a mock request with default URL options
@@ -398,7 +392,7 @@ class Form < ApplicationRecord
     )
 
     controller.request = mock_request
-    controller.render_to_string(partial: 'components/widget/widget', formats: :css, locals: { form: self })
+    controller.render_to_string(partial: 'components/widget/widget', formats: :css)
   end
 
   def reportable_submissions(start_date: nil, end_date: nil)
--- a/app/views/admin/questions/_question.html.erb
+++ b/app/views/admin/questions/_question.html.erb
@@ -9,39 +9,39 @@
   <% end %>
 </div>
 
-<%- if question.question_type == "text_field" %>
-<% @form_component_path = 'components/forms/edit/question_types/text_field' %>
-<% elsif question.question_type == "text_email_field" %>
-<% @form_component_path = 'components/forms/edit/question_types/text_email_field' %>
-<% elsif question.question_type == "text_phone_field" %>
-<% @form_component_path = 'components/forms/edit/question_types/text_phone_field' %>
-<% elsif question.question_type == "radio_buttons" %>
-<% @form_component_path = 'components/forms/edit/question_types/radio_buttons' %>
-<% elsif question.question_type == "big_thumbs_up_down_buttons" %>
-<% @form_component_path = 'components/forms/question_types/big_thumbs_up_down_buttons' %>
-<% elsif question.question_type == "dropdown" %>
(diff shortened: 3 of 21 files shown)


app/views/submissions/_form.html.erb
sanason, line 1: Moving the content of this partial into new.html.erb since that's the only place it's used.

app/views/components/forms/edit/_builder.html.erb
```diff
@@ -114,7 +114,7 @@
   <div class="grid-col-7">
     <br>
     <br>
```
sanason, line 116: This usage is within the form editor, not part of a deployed form.

app/controllers/touchpoints_controller.rb
```diff
@@ -13,7 +13,7 @@ def show
         js
       end
       format.css do
```
sanason, line 15: Assuming here that `/touchpoints/{:id}.css` should return the embedded version of the form CSS. I don't think this endpoint is actually used.

app/views/admin/questions/_question.html.erb
```diff
@@ -9,39 +9,39 @@
   <% end %>
 </div>
```
sanason, line 11: Switch to direct render for each question type so that different locals can be passed for different question types.

app/views/admin/questions/_question.html.erb
```diff
@@ -9,39 +9,39 @@
...
+<%- when "combobox" %>
+  <%= render 'components/forms/edit/question_types/combobox', question: question %>
+<%- when "textarea" %>
+  <%= render 'components/forms/edit/question_types/text_area', question: question %>
+<%- when "rich_textarea" %>
+  <%= render 'components/forms/edit/question_types/rich_textarea', question: question %>
+<%- when "hidden_field" %>
+  <%= render 'components/forms/edit/question_types/hidden_field', question: question %>
+<%- when "custom_text_display" %>
+  <%= render 'components/forms/question_types/custom_text_display', question: question, heading_level: 'h2' %>
```
sanason, line 38: This usage is within the form editor, not part of a deployed form.

app/models/form.rb
```diff
@@ -372,12 +372,6 @@ def touchpoints_js_string
     controller_with_request.render_to_string(partial: 'components/widget/fba', formats: :js, locals: { form: self })
   end
```
sanason, line 374: Delete a duplicatively-named method.

app/views/submissions/new.html.erb
```diff
@@ -1,21 +1,19 @@
 <div
   class="touchpoints-form-wrapper <%= @form.kind %>"
   data-touchpoints-form-id="<%= @form.short_uuid %>">
-<%
```
sanason, line 4: Remove this hack by adding a local variable to capture the context in which the form is being rendered instead of forcing that information into the form.

app/views/components/widget/_fba.js.erb
```diff
@@ -91,73 +127,46 @@ function FBAform(d, N) {
...
+        this.dialogEl = d.createElement('div');
+        this.dialogEl.setAttribute('class', "<%= form.prefix "usa-modal" %> fba-modal");
+        this.dialogEl.setAttribute('id', this.modalId());
+        this.dialogEl.setAttribute('aria-labelledby', `fba-form-title-${this.options.formId}`);
+        this.dialogEl.setAttribute('aria-describedby', `fba-form-instructions-${this.options.formId}`);
+        this.dialogEl.setAttribute('data-touchpoints-form-id', this.options.formId);
+
+        this.dialogEl.innerHTML = this.options.htmlFormBody();
+        d.body.appendChild(this.dialogEl);
+    }
```
sanason, line 156: Moved to `_bindEventListeners`.

